PR #3, opened 2020-10-19
Expose a listener API to get real time active message count

zhedahht:
Expose a listener API to get real time TopicEntity. The caller can get the useful information such as the count of active messages in the topic with the TopicEntity.

### DIFF
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -269,16 +269,16 @@ func (l *Listener) Close(ctx context.Context) error {
 }
 
 // GetActiveMessageCount gets the active message count of a topic subscription
+// WARNING: GetActiveMessageCount is 10 times expensive than a call to receive a message
 func (l *Listener) GetActiveMessageCount(ctx context.Context, topicName, subscriptionName string) (int32, error) {
-	topicEntity, err := getTopicEntity(ctx, topicName, l.namespace)
-	if err != nil {
-		return 0, fmt.Errorf("error to get entity of topic %q: %s", topicName, err)
+	if l.topicEntity == nil {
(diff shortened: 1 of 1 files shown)


listener/listener.go
```diff
@@ -267,6 +268,30 @@ func (l *Listener) Close(ctx context.Context) error {
 	return nil
 }
 
+// GetActiveMessageCount gets the active message count of a topic subscription
+func (l *Listener) GetActiveMessageCount(ctx context.Context, topicName, subscriptionName string) (int32, error) {
+	topicEntity, err := getTopicEntity(ctx, topicName, l.namespace)
```
paulgmiller, line 273: Since we do getTopicEnitty and getSubscription Entity is it two outgoing calls to arm? 
Asking since it will impact our thorttling if we hit arm twice. 

Can we reuse the topicentity from creation in getSubscriptionEntity or does it need to be a freshly pulled one?

listener/listener.go
```diff
@@ -267,6 +268,30 @@ func (l *Listener) Close(ctx context.Context) error {
 	return nil
 }
 
+// GetActiveMessageCount gets the active message count of a topic subscription
```
paulgmiller, line 271: Add a warning that this is at least 10x expsensive as a recieve call.

paulgmiller (approved):

listener/listener.go
zhedahht: Added a warning.

listener/listener.go
zhedahht: Yes, we don't really need a real time topic entity. Made the changes.

PR #4, opened 2020-10-23
Fix the issue that only the first message is handled

zhedahht:
The variable handler in listenerHandle comes from the outside. When the first message is handled, handler is changed, and it can't handle the following messages any more.

### DIFF
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -237,12 +237,12 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	// Create a handle class that has that function
 	listenerHandle := subReceiver.Listen(ctx, servicebus.HandlerFunc(
 		func(ctx context.Context, msg *servicebus.Message) error {
-			originalHandler := handler
-			for !message.IsDone(handler) {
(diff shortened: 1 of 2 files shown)


listener/listener.go
```diff
@@ -237,12 +237,12 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	// Create a handle class that has that function
 	listenerHandle := subReceiver.Listen(ctx, servicebus.HandlerFunc(
 		func(ctx context.Context, msg *servicebus.Message) error {
-			originalHandler := handler
-			for !message.IsDone(handler) {
-				handler = handler.Do(ctx, originalHandler, msg)
+			currentHandler := handler
```
paulgmiller, line 240: is there a unittest we can augment to catch this if we regress?

serbrech: Or write integration tests. They are easy to run too. The readme has a step by step to setup the environment using 2 make targets.

zhedahht: Added integration tests to publish and listen a message twice.

It failed without the fix, and passed when we have the fix.

PR #5, opened 2020-11-18
Add traces when handling messages

zhedahht:
Message handlers (abandon and complete don't return errors to the clients. Let's trace errors to facilitate troubleshooting.

We'd like to enable trace based on "github.com/devigned/tab", which is same as service bus.

### DIFF
--- a/message/abandon.go
+++ b/message/abandon.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	servicebus "github.com/Azure/azure-service-bus-go"
-	"github.com/devigned/tab"
+	"github.com/devigned/tab" // same way as service-bus to log errors
 )
 
 // Abandon stops processing the message and releases the lock on it.
@@ -20,7 +20,7 @@ func (a *abandon) Do(ctx context.Context, _ Handler, message *servicebus.Message
 	defer span.End()
 
 	if err := message.Abandon(ctx); err != nil {
-		tab.For(ctx).Debug(err.Error())
+		tab.For(ctx).Error(err)
 
 		// the processing will terminate and the lock on the message will be released after messageLockDuration
 	}
--- a/message/abandon.go
+++ b/message/abandon.go
@@ -20,7 +20,7 @@ func (a *abandon) Do(ctx context.Context, _ Handler, message *servicebus.Message
 	defer span.End()
 
 	if err := message.Abandon(ctx); err != nil {
-		span.AddAttributes(tab.StringAttribute("errorDetails", err.Error()))
+		tab.For(ctx).Debug(err.Error())
 
 		// the processing will terminate and the lock on the message will be released after messageLockDuration
 	}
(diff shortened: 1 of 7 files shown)


message/abandon.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	servicebus "github.com/Azure/azure-service-bus-go"
+	"github.com/devigned/tab"
```
paulgmiller, line 7: mention service bus is using this as well which is really nice since you can plugin your logger for both.

message/abandon.go
```diff
@@ -15,8 +16,12 @@ type abandon struct {
 }
 
 func (a *abandon) Do(ctx context.Context, _ Handler, message *servicebus.Message) Handler {
+	ctx, span := startSpanFromMessageAndContext(ctx, "go-shuttle.abandon.Do", message)
+	defer span.End()
+
 	if err := message.Abandon(ctx); err != nil {
-		// trace the failure to abandon the message.
+		tab.For(ctx).Debug(err.Error())
```
paulgmiller, line 23: Why debug? Why not Error(err.Error()) warning.

Same for complete below.

message/error.go
```diff
@@ -15,7 +16,13 @@ type errorHandler struct {
 	err error
 }
 
-func (a *errorHandler) Do(_ context.Context, _ Handler, _ *servicebus.Message) Handler {
-	// can trace the error before abandoning once we settle on a tracing system that can be passed via context
+func (a *errorHandler) Do(ctx context.Context, _ Handler, msg *servicebus.Message) Handler {
+	ctx, span := startSpanFromMessageAndContext(ctx, "go-shuttle.errorHandler.Do", msg)
+	defer span.End()
+
+	if a.err != nil {
```
paulgmiller, line 23: unlike the ones above this is less important to log since it comes from the user and they can log it if they want but doesn't hurt to double log probably.

message/tracing.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	servicebus "github.com/Azure/azure-service-bus-go"
+	"github.com/devigned/tab"
+)
+
+func startSpanFromMessageAndContext(ctx context.Context, operationName string, message *servicebus.Message) (context.Context, tab.Spanner) {
+	ctx, span := tab.StartSpan(ctx, operationName)
+	if message != nil {
```
paulgmiller, line 12: this should never happen. Log an erro when it does?

paulgmiller (approved):

message/abandon.go
zhedahht: mentioned in comments

message/abandon.go
zhedahht: changed to Error

message/error.go
zhedahht: It makes sense. It's not necessary to log the error here.

message/tracing.go
zhedahht: Log an error when msg is nil.

PR #7, opened 2020-12-09
Expose delete topic in publisher

xhl873:


publisher/publisher.go
```diff
@@ -212,3 +212,22 @@ func ensureTopic(ctx context.Context, name string, namespace *servicebus.Namespa
 
 	return tm.Put(ctx, name, opts...)
 }
+
+// Delete the given topic
+func DeleteTopic(topicName string, opts ...ManagementOption) error {
+	ns, err := servicebus.NewNamespace()
```
paulgmiller, line 218: nit factor out into helper shared with New?

paulgmiller (approved):
seems pretty straight forward. I'm assuming you can check the dedup window

xhl873:
> seems pretty straight forward. I'm assuming you can check the dedup window and delete if it differs from what you want.

we could check the dedup window in go-shuttle, but not outside, the topic entity is also not exposed. I haven't try to get the topic entity to check the properties, it should be doable in publisher code. was thinking about only delete/recreate the topic when we explicitly say we want to do that on our side.

PR #9, opened 2021-01-04
strawman to show how we might do concurrency

paulgmiller:


paulgmiller:
@serbrech  is probably on a plane right now but should be intereste in this since I asssume his mental model was the same as mine that each message got its own go routine. Otherwise retry later holding the message makes no sense.

listener/listener.go
```diff
@@ -234,19 +243,36 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
...
-					currentHandler = message.Complete()
+	//blantantly stolen from https://github.com/Azure/azure-service-bus-go/pull/117/files
+	msgChan := make(chan *servicebus.Message, l.concurrency)
+	// Define a function that should be executed when a message is received.
+	var concurrentHandler servicebus.HandlerFunc = func(ctx context.Context, msg *servicebus.Message) error {
+		msgChan <- msg
+		return nil
+	}
+	if l.concurrency == 0 {
+		l.concurrency = 5 //what's the right default? 1? something higher?
```
paulgmiller, line 254: this actually needs to be pretty high or retry later will still sink us.

paulgmiller: We could remove any max concurrancy here and allways spin up as many go functions as we have messages. That is closer to what charlie was considering.

integration/pubsub_test.go
```diff
@@ -227,6 +254,93 @@ func (suite *serviceBusSuite) duplicateDetectionTest(pub *publisher.Publisher, l
...
+		testConfig.listenerOptions = []listener.Option{}
+	}
+	if testConfig.publisherOptions == nil {
+		testConfig.publisherOptions = []publisher.Option{}
+	}
+
+	events := []testEvent{}
+	for i := 0; i < numOfEvents; i++ {
+		j := i + 1
+		event := testEvent{
```
paulgmiller, line 269: this could just be anonymous empty struct we're not actually using the data here.

integration/pubsub_test.go
```diff
@@ -227,6 +254,93 @@ func (suite *serviceBusSuite) duplicateDetectionTest(pub *publisher.Publisher, l
...
+			}
+		}
+		if len(msgstatus) >= len(expected) {
+			break
+		}
+	}
+
+	assert.Equal(suite.T(), msgstatus, expected)
+
+	//time.Sleep(20 * time.Second)
```
xhl873, line 339: this is added to not close listener before it prints out the msgs, I think it's not need anymore as for the loop above blocks until we finish the checks.

integration/pubsub_test.go
```diff
@@ -114,6 +117,30 @@ func (suite *serviceBusSuite) TestPublishAndListenRetryLater() {
 	}, event)
 }
 
+// Note that, this test need manually verify, you should be able to see 3 msgs are receievd in a row from the test log
```
xhl873, line 120: it's awesome to have the order verification, the comment here is outdated :D

xhl873 (approved):

integration/pubsub_test.go
```diff
@@ -227,6 +254,93 @@ func (suite *serviceBusSuite) duplicateDetectionTest(pub *publisher.Publisher, l
...
+		err := testConfig.publisher.Publish(
+			ctx,
+			events[i],
+			testConfig.publisherOptions...,
+		)
+		suite.NoError(err)
+	}
+	var msgstatus []string
+	expected := []string{
+		"recieved1",
```
paulgmiller, line 305: I don't think 123 are actually gartuneed in that order.

integration/pubsub_test.go
serbrech: question, why do we have order verification?
servicebus can't guarantee order by design. especially with concurrent workers pulling from the subscription?

listener/listener.go
```diff
@@ -224,8 +233,12 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 		_ = topic.Close(ctx)
 	}()
 
+	if l.concurrency == 0 {
+		l.concurrency = 1 //RetryLater won't work with this. Come back and change after talking to strebec?
+	}
+	prefetch := servicebus.SubscriptionWithPrefetchCount(uint32(l.concurrency))
```
serbrech, line 239: I don't think prefetch is strictly like to concurrency.
prefetch will make the request fetch N messages at once. that's much better obviously.
but since the messages can be processed asynchronously from the fetching, it's not a requirement.

- fetch message1
- start process message 1
- fetch message2
- start process message 2
- done process message 2
- done process message 1

the above is concurrent processing with a prefetch of 1.

- fetch message1 and message2
- start process message 1
- start process message 2
- done process message 2
- done process message 1

The above is concurrent processing with prefetch of 2 (assuming the messages are present in the subscription when the receiver starts)

integration/pubsub_test.go
paulgmiller: Really we wanted to show that we recieved messages before we completed others (actually proved we were parallel) Showing the ordering was mostly for debugging the test could just be that we got a certain number of recieves before completes.

listener/listener.go
paulgmiller: Agree it's not strictly tied though probably a good default.
Not sure this PR is useful to autoupgrader or replacing the rp without lock auto-renewal.   Take a look at message pump here to see concurrency and auto renewal. I kind of want to get that up to the servicebus sdk as an optional handler any consumer can use. 

https://github.com/Azure/go-shuttle/pull/13/files#diff-f8191b0240dc42b4c68b583cf4f8c33a1c63d8bd3537410acdb0f142a25ec939

PR #11, opened 2021-01-06
Retrylater should renew the lock

xhl873:
draft pr for possible solution to renewlock in retrylater

### DIFF
--- a/message/retrylater.go
+++ b/message/retrylater.go
@@ -8,6 +8,8 @@ import (
 	servicebus "github.com/Azure/azure-service-bus-go"
 )
 
+const renewLockInterval = 20 * time.Second
+
 // RetryLater waits for the given duration before retrying the processing of the message.
 // This happens in memory and does not impact servicebus message max retry limit
 func RetryLater(retryAfter time.Duration) Handler {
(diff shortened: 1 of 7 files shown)


message/retrylater.go
```diff
@@ -20,11 +20,12 @@ type retryLaterHandler struct {
 	retryAfter time.Duration
 }
 
-func (r *retryLaterHandler) Do(ctx context.Context, orig Handler, message *servicebus.Message) Handler {
+func (r *retryLaterHandler) Do(ctx context.Context, orig Handler, message *servicebus.Message, sub *servicebus.Subscription) Handler {
 	select {
 	case <-ctx.Done():
 		return Error(fmt.Errorf("aborting retrylater for message. abandoning meddage %s: %w", message.ID, ctx.Err()))
 	case <-time.After(r.retryAfter):
```
paulgmiller, line 27: if retryafter is longer than lockduration you have a problem. Maybe add a third case and renew after either peek lock or some arbitrary const for now

message/retrylater.go
```diff
@@ -20,11 +20,12 @@ type retryLaterHandler struct {
...
+		sub.RenewLocks(ctx, message)
```
paulgmiller, line 28: does renew locks have a return code?

serbrech:
I think I prefer Paul's attempt that moves this higher up in the message handling code.
the Lock should always be renewed until the message is : 
- abandoned
- completed
- an error occurs
- we've reached a processing timeout or a maxlock timeout

thoughts?

xhl873:
> I think I prefer @paulgmiller attempt that moves this higher up in the message handling code.
> the Lock should always be renewed until the message is :
> 
> * abandoned
> * completed
> * an error occurs
> * we've reached a processing timeout or a maxlock timeout
> 
> thoughts?

Yes, totally agree with you. We were exploring the possibilities at that time, and when I test this solution, found out that some of the msgs already expired before land on the handler. You are right, I think's the lock renew should be in higher levels

serbrech:
closing now that #16 is merged

PR #13, opened 2021-01-06
Pmiller/renewlocks

paulgmiller:


listener/listener.go
```diff
@@ -247,29 +248,40 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 		return fmt.Errorf("failed to create new subscription receiver %s: %w", l.subscriptionEntity.Name, err)
 	}
 
-	//blantantly stolen from https://github.com/Azure/azure-service-bus-go/pull/117/files
-	msgChan := make(chan *servicebus.Message, l.concurrency)
-	// Define a function that should be executed when a message is received.
+	semaphore := make(chan bool, l.concurrency)
```
charliedmcb, line 251: I don't think the semaphore is actually necessary as I believe it the listener can only get the number of messages equal to prefetch at a time. If it is necessary to block than I think there is an issue as the message will have already been accepted, so the lock will be ticking down even though it is not ready to handle the message yet.

serbrech:
clsoing as #16 is merged

PR #14, opened 2021-01-07
RetryLater should abandon on a seperate go routine

paulgmiller:
Right now retry later is broken because it will 
1) Block other messages from coming in
2) Block till after lock duration which means completes error  and messages don't finish

### DIFF
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -131,15 +131,21 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
 	}
 }
 
-//allows listeners to control subscription details for longer lived operations.
-//If you using RetryLater you probably want this
-func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) Option {
+// WithSubscriptionDetails allows listeners to control subscription details for longer lived operations.
+// If you using RetryLater you probably want this. Passing zeros leaves it up to Service bus defaults
+func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) ManagementOption {
 	return func(l *Listener) error {
 		if lock > servicebusinternal.LockDuration {
 			//working on getting service bus to enforce this. Hangs if you go higher. https://github.com/Azure/azure-service-bus-go/pull/202
 			return fmt.Errorf("Lock duration must be <= to %v", servicebusinternal.LockDuration)
 		}
+		if lock < time.Duration(0) {
+			return fmt.Errorf("Lock duration must be positive")
+		}
 		l.lockDuration = lock
+		if maxDelivery < 0 {
+			return fmt.Errorf("Max Deliveries must be positive")
+		}
 		l.maxDeliveryCount = maxDelivery
 		return nil
 	}
@@ -189,7 +195,7 @@ func setSubscriptionEntity(ctx context.Context, l *Listener) error {
 	if l.subscriptionName == "" {
 		l.subscriptionName = defaultSubscriptionName
 	}
-	subscriptionEntity, err := l.getSubscriptionEntity(ctx, l.subscriptionName, l.namespace, l.topicEntity)
+	subscriptionEntity, err := l.getSubscriptionEntity(ctx, l.subscriptionName)
 	if err != nil {
 		return fmt.Errorf("failed to get subscription: %w", err)
 	}
@@ -295,7 +301,7 @@ func (l *Listener) GetActiveMessageCount(ctx context.Context, topicName, subscri
 		return 0, fmt.Errorf("topic name %q doesn't match %q", topicName, l.topicName)
 	}
 
-	subscriptionEntity, err := l.getSubscriptionEntity(ctx, subscriptionName, l.namespace, l.topicEntity)
+	subscriptionEntity, err := l.getSubscriptionEntity(ctx, subscriptionName)
 	if err != nil {
 		return 0, fmt.Errorf("error to get entity of subscription %q of topic %q: %s", subscriptionName, topicName, err)
 	}
@@ -316,10 +322,9 @@ func getTopicEntity(ctx context.Context, topicName string, namespace *servicebus
 
 func (l *Listener) getSubscriptionEntity(
 	ctx context.Context,
-	subscriptionName string,
-	ns *servicebus.Namespace,
-	te *servicebus.TopicEntity) (*servicebus.SubscriptionEntity, error) {
-	subscriptionManager, err := ns.NewSubscriptionManager(te.Name)
+	subscriptionName string) (*servicebus.SubscriptionEntity, error) {
+
+	subscriptionManager, err := l.namespace.NewSubscriptionManager(l.topicEntity.Name)
 	if err != nil {
 		return nil, fmt.Errorf("creating subscription manager failed: %w", err)
 	}
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -338,8 +338,13 @@ func (l *Listener) ensureSubscription(ctx context.Context, sm *servicebus.Subscr
 		return subEntity, nil
 	}
 	mutateSericeDetails := func(s *servicebus.SubscriptionDescription) error {
-		s.MaxDeliveryCount = &l.maxDeliveryCount
-		return servicebus.SubscriptionWithLockDuration(&l.lockDuration)(s)
+		if l.maxDeliveryCount != 0 {
+			s.MaxDeliveryCount = &l.maxDeliveryCount
+		}
+		if l.lockDuration > time.Duration(0) {
+			return servicebus.SubscriptionWithLockDuration(&l.lockDuration)(s)
+		}
+		return nil
 	}
 	return sm.Put(ctx, name, mutateSericeDetails)
 }
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"time"
 
 	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/Azure/go-autorest/autorest/adal"
@@ -16,7 +15,6 @@ import (
 
 const (
 	defaultSubscriptionName = "default"
-	LockDuration            = 5 * time.Minute //Need to reference from retry later.
 )
 
 // Listener is a struct to contain service bus entities relevant to subscribing to a publisher topic
@@ -327,8 +325,7 @@ func ensureSubscription(ctx context.Context, sm *servicebus.SubscriptionManager,
 		s.MaxDeliveryCount = &maxDeliveryCount //whats our biggest agentpool? this is 5 * 500 minutes
 		return nil
 	}
-
-	duration := LockDuration
+	duration := servicebusinternal.LockDuration
 	return sm.Put(ctx, name, servicebus.SubscriptionWithLockDuration(&duration), increaseMaxDelivery)
 }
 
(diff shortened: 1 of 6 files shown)


listener/listener.go
```diff
@@ -320,8 +322,14 @@ func ensureSubscription(ctx context.Context, sm *servicebus.SubscriptionManager,
 	if err == nil {
```
xhl873, line 320: we are returning it once the subscription is already there, this is sad when we want to do some management operations. it'll be good if "sm.Put" can update the setting, we can let user choose when to update, Otherwise may need to find some way to clean the existing subscriptions.

message/retrylater.go
```diff
@@ -21,10 +20,13 @@ type retryLaterHandler struct {
 }
 
 func (r *retryLaterHandler) Do(ctx context.Context, orig Handler, message *servicebus.Message) Handler {
-	select {
-	case <-ctx.Done():
-		return Error(fmt.Errorf("aborting retrylater for message. abandoning meddage %s: %w", message.ID, ctx.Err()))
-	case <-time.After(r.retryAfter):
-		return orig
-	}
+	go func() {
```
xhl873, line 24: is the new thread there to make it return faster? but then we may lose the ctx timeout error.

listener/listener.go
```diff
@@ -320,8 +320,13 @@ func ensureSubscription(ctx context.Context, sm *servicebus.SubscriptionManager,
...
 		return subEntity, nil
 	}
-
-	return sm.Put(ctx, name)
+	increaseMaxDelivery := func(s *servicebus.SubscriptionDescription) error {
+		var maxDeliveryCount int32 = 500
```
zhedahht, line 324: Do we increase the delivery count for every message? We have a few cases to abandon messages. If the answer is yes, does it mean that every message will be abandoned much more times?

message/retrylater.go
paulgmiller: yes the go function means we don't block other items but it does make it harder to timeout with conext. Not sure we shoudl even check context.

listener/listener.go
paulgmiller: Yes this means messages that go through retry laters will be aboned/delivered several times. And if we go over 500*5 minutes then we give up but I think that's okay as long as its higher than when regional looper will give us a new message.

message/retrylater.go
charliedmcb: Correct me if I'm wrong, but I don't think returning early here is necessarily useful, and possibly not actually a good idea.

If I understand it correctly, the message will still be locked on the topic until the timeout and the message is abandoned, so will not be able to pick up a new message until the time elapses anyways.

Plus, in my digging I spotted what seem to be a setting in azure-service-bus-go where the handleMessage function within side of azure-service-bus-go will handle the Disposition. I don't completely understand how the Disposition code works, but it makes me think that it could be possible that the thread is waiting here to abandon, but by returning the listener actually disposes of the msg from the topic. I don't believe this is currently being used by go-shuttle or exposed as a setting, but by returning here instead of letting the Abandon play out it seems like it could result in unexpected behavior down the line if more changes are made to the repo.

listener/listener.go
charliedmcb, line 324: I don't think hardcoding this as the final solution is a good idea. If we take this approach I think we should expose setting maxDeliveryCount as a setting to the user

message/retrylater.go
charliedmcb: From the most recent test I ran I think my assumption here about how the locking was incorrect. I detailed why I think my assumption was wrong in the chat.

message/retrylater.go
```diff
@@ -21,10 +21,19 @@ type retryLaterHandler struct {
...
+		//TODO this can go past lock duration pretty easily
+		//Ideally we'd also timeout at dequeue time + lockdurtaion - 1 second but don't have access to dequeue time
+		//Maybe we can use context.WithTimeout when we recieve the message?
+		case <-ctx.Done():
+			//listener should cancel before lock duration which is the longest we can ask for retry later
+			fmt.Printf("messsage context canceled\n")
+			return
+		case <-time.After(r.retryAfter):
+			fmt.Printf("Abandoning messsage %v\n", time.Now())
+			Abandon().Do(ctx, orig, message)
```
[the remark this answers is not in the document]
paulgmiller: good catch updated

listener/listener.go
paulgmiller: Made configurable.

listener/listener.go
paulgmiller: Agree though punting to a different change. For now would have to change subscription name,

listener/listener.go
```diff
@@ -128,6 +131,20 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
 	}
 }
 
+//allos listeners to control subscription details for longer lived operations.
```
xhl873, line 134: typo

listener/listener.go
```diff
@@ -128,6 +131,20 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
...
+//If you using RetryLater you probably want this
+func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) Option {
```
xhl873, line 136: this need to be ManagementOption in order to be used when New the listener

xhl873 (approved):

listener/listener.go
```diff
@@ -307,21 +324,29 @@ func getSubscriptionEntity(
...
-func ensureSubscription(ctx context.Context, sm *servicebus.SubscriptionManager, name string) (*servicebus.SubscriptionEntity, error) {
+func (l *Listener) ensureSubscription(ctx context.Context, sm *servicebus.SubscriptionManager, name string) (*servicebus.SubscriptionEntity, error) {
 	subEntity, err := sm.Get(ctx, name)
 	if err == nil {
 		return subEntity, nil
 	}
-
-	return sm.Put(ctx, name)
+	mutateSericeDetails := func(s *servicebus.SubscriptionDescription) error {
+		if l.maxDeliveryCount != 0 {
```
charliedmcb, line 341: nit: > 0
```suggestion
		if l.maxDeliveryCount > 0 {
```

listener/listener.go
```diff
@@ -128,6 +131,20 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
...
 
+//allos listeners to control subscription details for longer lived operations.
+//If you using RetryLater you probably want this
+func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) Option {
+	return func(l *Listener) error {
+		if lock > servicebusinternal.LockDuration {
+			//working on getting service bus to enforce this. Hangs if you go higher. https://github.com/Azure/azure-service-bus-go/pull/202
+			return fmt.Errorf("Lock duration must be <= to %v", servicebusinternal.LockDuration)
+		}
+		l.lockDuration = lock
```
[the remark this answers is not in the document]
charliedmcb: I see you do the check where the value is used

listener/listener.go
```diff
@@ -128,6 +131,20 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
...
+		l.maxDeliveryCount = maxDelivery
```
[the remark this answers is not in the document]
charliedmcb: I see you do the check where the value is used, but need to check > 0, left a suggestion there

message/retrylater.go
```diff
@@ -2,14 +2,15 @@ package message
...
 
 	servicebus "github.com/Azure/azure-service-bus-go"
+	"github.com/devigned/tab"
 )
 
-// RetryLater waits for the given duration before retrying the processing of the message.
-// This happens in memory and does not impact servicebus message max retry limit
+// RetryLater waits for the given duration before abanoning.
+// Consider examining/increasing your LockDuration/MaxDeliveryCount before using.
+// Can't retryly later longer than the lock duration (which has a max of 5 minutes)
```
charliedmcb, line 13: nit: Change to something like "Undefined behavior if RetryLater is passed a duration that puts the message handling past the lock duration (which has a max of 5 minutes)"

If you like the suggestion:
```suggestion
// Undefined behavior if RetryLater is passed a duration that puts the message handling past the lock duration (which has a max of 5 minutes)
```

message/retrylater.go
```diff
@@ -21,10 +22,19 @@ type retryLaterHandler struct {
...
-		return Error(fmt.Errorf("aborting retrylater for message. abandoning meddage %s: %w", message.ID, ctx.Err()))
-	case <-time.After(r.retryAfter):
-		return orig
-	}
+	go func() {
+		ctx, span := startSpanFromMessageAndContext(ctx, "go-shuttle.retryLater.Do", message)
+		defer span.End()
+
+		select {
+		//TODO this can go past lock duration pretty easily
```
charliedmcb, line 30: Feels concerning to leave this, but I think its ok to complete. Its much better than before :)

listener/listener.go
```diff
@@ -172,7 +189,7 @@ func setSubscriptionEntity(ctx context.Context, l *Listener) error {
 	if l.subscriptionName == "" {
 		l.subscriptionName = defaultSubscriptionName
 	}
-	subscriptionEntity, err := getSubscriptionEntity(ctx, l.subscriptionName, l.namespace, l.topicEntity)
+	subscriptionEntity, err := l.getSubscriptionEntity(ctx, l.subscriptionName, l.namespace, l.topicEntity)
```
charliedmcb, line 192: I noticed in both of the calls to this function l.namespace, and l.topicEntity were passes in. Do you expect the values will always come from the listener when calling this function, if so seems like could be retrieved from the listener struct it is being called on

charliedmcb (approved):
Mostly nits
Still concerned about RetryLater taking longer than lock duration, but I think its ok to complete.

listener/listener.go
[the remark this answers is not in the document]
charliedmcb: Still seems better to do the check here, so no invalid data is stored

listener/listener.go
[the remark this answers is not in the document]
charliedmcb: Still seems better to do the check here, so no invalid data is stored

message/retrylater.go
paulgmiller: Agree. Lets create an issue on the github too. I think for RP we want to go back to a model where we constantly renew in the background.

listener/listener.go
paulgmiller: I don't think it mattters if we set it at New or at Listen but changed to managedment option.

listener/listener.go
```diff
@@ -128,6 +131,20 @@ func WithFilterDescriber(filterName string, filter servicebus.FilterDescriber) M
 	}
 }
 
+//allos listeners to control subscription details for longer lived operations.
+//If you using RetryLater you probably want this
```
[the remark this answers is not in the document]
paulgmiller: did so.

listener/listener.go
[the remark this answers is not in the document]
paulgmiller: Added

listener/listener.go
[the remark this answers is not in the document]
paulgmiller: Added

listener/listener.go
paulgmiller: yep changed that.

listener/listener.go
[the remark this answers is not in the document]
charliedmcb: I saw you had something there, but I was actually meaning adding this info in addition to what you already have. I think it would be good if someone was going to use RetryLater that they were directed to also set WithSubscriptionDetails, especially maxDelivery

PR #15, opened 2021-01-08
Updating README.md

charliedmcb:
I noticed that after the recent change to RetryLater that there should be some additional updates to the README for the explination of RetryLater. These are those changes :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ err := l.Listen(ctx, handler, topicName)
 ```
 
 Notes: 
-* RetryLater simply waits for the given duration before abanoning. So, if you using RetryLater you probably want to set WithSubscriptionDetails, especially maxDelivery as each call to RetryLater will up the delivery count by 1
+* RetryLater simply waits for the given duration before abanoning. So, if you use RetryLater you probably want to set WithSubscriptionDetails, especially maxDelivery as each call to RetryLater will up the delivery count by 1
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,17 +99,17 @@ err := l.Listen(ctx, handler, topicName)
...
-
 // listen blocks and handle messages from the topic
 err := l.Listen(ctx, handler, topicName)
 ```
 
-Note that this happens in-memory in your service. The receiver is keeping your message and pushing it back to your handler after the given time.
-This will not increase the retry count of the message, as the message is not dequeued another time.
+Notes: 
+* RetryLater simply waits for the given duration before abanoning. So, if you are using RetryLater you probably want to set WithSubscriptionDetails, especially maxDelivery as each call to RetryLater will up the delivery count by 1
+* Undefined behavior if RetryLater is passed a duration that puts the message handling (time used in go-shuttle, client handler, and RetryLater combined) past the lock duration (which has a max of 5 minutes).
```
paulgmiller, line 112: Undefined is a little strong. Your abandon should just error out and your later won't be as late but I'm okay being a little scary.

paulgmiller (approved):

README.md
charliedmcb: I was thinking it would be best to have it be a bit scary as I thought it will currently try to Abandon the message without a lock which I expect will leave the message in the non-completed/completed state where it doesn't get removed from the topic, but also never handled again or am I wrong?

PR #16, opened 2021-01-22
Add peeklock auto renewal functionality

serbrech:
This adds automatic Peek Lock renewal functionality to the message listener.

```golang
listener.WithMessageLockAutoRenewal(interval time.Duration) Option {
...
}
```

The above listener Options allows you to setup auto-renewal interval : 

```golang
err := l.Listen(
    ctx,
    handler,
    topicName,
    listener.MessageLockAutoRenewal(30 * time.Second),
)
```

This does not look at message expiration and only takes care of renewal of the lock.
It stops renewing when the message is completed or abandoned, but will not stop renewing if the message is expired.
Message Expiration handling is to be implemented in a separate PR


serbrech:
Message Expiration handling issue : #17

peeklock/renewer.go
```diff
@@ -0,0 +1,54 @@
...
+	return p
+}
+
+func (plr *PeriodicLockRenewer) startPeriodicRenewal(ctx context.Context, interval time.Duration, message *servicebus.Message) {
+	var renewerCtx context.Context
+	renewerCtx, plr.cancelFunc = context.WithCancel(ctx)
+	for alive := true; alive; {
+		select {
+		case <-time.After(interval):
+			// add span for logs
```
xhl873, line 34: would u like to uncomment the logs? :D

peeklock/renewer.go
```diff
@@ -0,0 +1,54 @@
...
+			// fmt.Printf("%s - Renewing Lock: [ ttl: %s, lockedUntil: %s, enqueuedTime: %s ]",
+			// 	time.Now(),
+			// 	message.TTL,
+			// 	message.SystemProperties.LockedUntil,
+			// 	message.SystemProperties.EnqueuedTime)
+			plr.lockRenewer.RenewLocks(renewerCtx, message)
+			// what do we do if we fail to renew?
```
xhl873, line 41: will it be too aggressive to discard the msg if renew lock failed? the msg might be picked up by another thread if we still keep it after lock renew failed

another question might be more related to ttl, I am curious if we trying to complete a msg after ttl, will it simply failed or even blocked the thread?

peeklock/renewer.go
```diff
@@ -0,0 +1,54 @@
...
+			alive = false
+		}
+	}
+}
+
+func (plr *PeriodicLockRenewer) Stop() {
+	if plr.cancelFunc != nil {
+		plr.cancelFunc()
+	}
+
```
xhl873, line 53: nit: extra newline

peeklock/renewer.go
serbrech: No that's don't print. I added an issue to move the tracing helper I to a package, then we can trace everywhere. That was for test. I'll delete the commented logs

peeklock/renewer.go
serbrech: I think that's a good idea. If we fail to renew the lock, just fail
The message.
The underlying rpc call has retries already, so if we fail here, something's wrong. Thanks 👍

integration/pubsub_test.go
```diff
@@ -401,13 +471,26 @@ func checkResultHandler(publishedMsg string, publishedMsgType string, ch chan<-
...
 					return message.Complete()
 				}
 				return message.RetryLater(1 * time.Second)
 			}
+			if publishedMsgType == reflection.GetType(shortLockMessage{}) {
+				// the shortlock is set at 1 second in the listener setup
+				// We sleep for longer to trigger the lock renewal in the listener.
+				// if the renewal works, then the message completion will succeed.
+				time.Sleep(3 * time.Second)
+				res := message.Complete()
```
paulgmiller, line 486: if renewal wasn't working we'd fail here or would renewal cancel this? Worth an integration test to show we fail if we lost the lock?

integration/pubsub_test.go
```diff
@@ -114,6 +114,33 @@ func (suite *serviceBusSuite) TestPublishAndListenRetryLater() {
...
+func (suite *serviceBusSuite) TestPublishAndListenShortLockDuration() {
+	suite.T().Parallel()
+	// creating a separate topic that was not created at the beginning of the test suite
+	// note that this topic will also be deleted at the tear down of the suite due to the tagID at the end of the topic name
+	shortLockTopic := suite.Prefix + "shortlock" + suite.TagID
+	pub, err := publisher.New(shortLockTopic, suite.publisherAuthOption)
+	suite.NoError(err)
+	l, err := listener.New(
+		suite.listenerAuthOption,
+		listener.WithSubscriptionDetails(2*time.Second, 3),
```
paulgmiller, line 126: Sigh my WithSubscriptionDetails isn't very explicit about what these values are setting. My fault.

integration/pubsub_test.go
```diff
@@ -401,13 +471,26 @@ func checkResultHandler(publishedMsg string, publishedMsgType string, ch chan<-
...
 					ch <- true
 					return message.Complete()
 				}
 				return message.RetryLater(1 * time.Second)
 			}
+			if publishedMsgType == reflection.GetType(shortLockMessage{}) {
+				// the shortlock is set at 1 second in the listener setup
+				// We sleep for longer to trigger the lock renewal in the listener.
+				// if the renewal works, then the message completion will succeed.
+				time.Sleep(3 * time.Second)
```
paulgmiller, line 485: crank this up for at least one test and see if we can see any of the renew failures  that @xhl873  saw?

listener/listener.go
```diff
@@ -151,6 +153,16 @@ func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) ManagementOp
 	}
 }
 
+func WithMessageLockAutoRenewal(interval time.Duration) Option {
+	return func(l *Listener) error {
+		if interval < time.Duration(0) {
+			return fmt.Errorf("renewal interval must be positive")
```
paulgmiller, line 159: what happens if I set it to 0? Should we have a lower limit of a second to stop sillyness?

listener/listener.go
```diff
@@ -260,6 +272,10 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	// Create a handle class that has that function
 	listenerHandle := subReceiver.Listen(ctx, servicebus.HandlerFunc(
 		func(ctx context.Context, msg *servicebus.Message) error {
+			if l.lockRenewalInterval != nil {
+				renewer := peeklock.RenewPeriodically(ctx, *l.lockRenewalInterval, sub, msg)
+				defer renewer.Stop()
```
paulgmiller, line 277: why the stop idiom rather than doing  
```
renewerecontext, cancel = context.WithCancelation(ctx)
defer cancel()
peeklock.RenewPeriodically(renewercontext)
```
My understanding is that the cancel only cancels renewercontext and not ctx.

peeklock/renewer.go
paulgmiller: Ah renewlocks already has retries? Do you know how much it backs off? My original thought was to keep retrying till we got to lock duration but internal retries makes that dubious

My prefernce if we fail after retries is to cancel the cotnext we gave to the user provideded handler so they can stop if need be.

listener/listener.go
paulgmiller: ah you are doing that inside renwew lock but why return a wholre renewer rather than just a cancel function. Maybe you're thinkiing of renewer having something else eventually?

paulgmiller (approved):
Mostly curious about canelation of worker if we fail to renew.

peeklock/renewer.go
serbrech: under the hood, renewLocks does : 
```
client, err := re.entity.GetRPCClient(ctx)
	if err != nil {
		tab.For(ctx).Error(err)
		return err
	}
	return client.RenewLocks(ctx, messages...)
```

the rpcclient always does (minus error handling for simplicity) : 

```
if err := r.ensureConn(ctx); err != nil {
		tab.For(ctx).Error(err)
		return err
	}
...
rpcLink, err := rpc.NewLink(r.client, r.ec.ManagementPath())
... 
response, err := rpcLink.RetryableRPC(ctx, 3, 1*time.Second, renewRequestMsg)
...
```

serbrech: with retryableRPC signature being: 

```
func (l *Link) RetryableRPC(ctx context.Context, times int, delay time.Duration, msg *amqp.Message) (*Response, error) {
```

integration/pubsub_test.go
serbrech: yes, I say it derived a little from the pattern. not a big deal, but we can fix it in a future PR, and break ALL OUR USERS ;)

listener/listener.go
serbrech: it will just keep looping on renewal, probably causing 429 really quick.
I think the flogging is equal to the silliness.

listener/listener.go
serbrech: I started with the struct, and I got there this way.
Plan for the renewer would be to be able to batch concurrent message renewal to avoid increasing the number of calls too fast when increasing concurrency. I'm just not sure how it will look like. since this context is created inside the renewer, I thought it was cleaner to keep it private.

integration/pubsub_test.go
serbrech: yes, it will fail Completing the message.
Once we have a per message context, we can cancel that context.
the behavior of the message handling will be the same though, since canceling isn't stopping processing, it's just preventing future calls that use the context.
The improvement will be that we'll fail on our side rather than after the RPC call.
so, leaving it as is until we have per message contexts.

integration/pubsub_test.go
serbrech: Will try. I think what @xhl873 saw was message expiration, which prevents any future actions on the message (TTL).
to be confirmed.

peeklock/renewer.go
serbrech: running go fmt ADDS this end of tile line :)

listener/listener.go
```diff
@@ -151,6 +153,16 @@ func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) ManagementOp
 	}
 }
 
+func WithMessageLockAutoRenewal(interval time.Duration) Option {
```
charliedmcb, line 156: If the interval is greater than lock duration than it will simply do nothing, correct? Seems like ok behavior, but thought I'd call it out in case we want to call it out in a comment.

serbrech: that's correct. if the lock expire after 10 sec, and you auto-renew every 20 second, you're not doing anything.
leaving this as a potential improvement, and good enough for now.

serbrech:
After the fix in go-amqp to bubble the error to the receiver, we can see the error showing up in the traces 👍 

jaeger

.gitignore
```diff
@@ -14,5 +14,6 @@
 # Dependency directories (remove the comment below to include it)
 # vendor/
 .env
+test-debug.env
```
serbrech, line 17: ignoring this file as I used it to run integration test locally and needed more than just .env (goland wants a visible file...)

Makefile
```diff
@@ -1,5 +1,5 @@
 include .env
-IMAGE?="${REGISTRY}/pubsubtest"
+IMAGE?=${REGISTRY}/pubsubtest
```
serbrech, line 2: docker-compose did not like the quotes, and they are superfluous

Makefile
```diff
@@ -44,5 +44,7 @@ clean-aci:
 integration: build-test-image push-test-image test-aci
 
 integration-local: build-test-image
-	@docker run --env-file .env --rm --entrypoint go test -run TestConnectionString -v ./integration ${IMAGE}
+	@docker-compose --env-file .env up
```
serbrech, line 47: added a docker-compose.yaml file that sets up jaeger and runs integration tests locally (targetting the servicebus namespace as configured in .env)

go.mod
```diff
@@ -7,9 +7,17 @@ require (
...
 	github.com/onsi/gomega v1.10.2
+	github.com/opentracing/opentracing-go v1.2.0
 	github.com/stretchr/testify v1.6.1
+	github.com/uber/jaeger-client-go v2.25.0+incompatible
+	github.com/uber/jaeger-lib v2.4.0+incompatible
+	go.uber.org/atomic v1.7.0 // indirect
 	golang.org/x/crypto v0.0.0-20200728195943-123391ffb6de
 )
+
+replace github.com/Azure/go-amqp => github.com/serbrech/go-amqp fix-disposition-error
```
serbrech, line 23: using my branch until it merges to master in `go-amqp`

integration/pubsub_test.go
```diff
@@ -114,6 +117,60 @@ func (suite *serviceBusSuite) TestPublishAndListenRetryLater() {
...
+	suite.publishAndReceiveMessageWithAutoLockRenewal(publishReceiveTest{
+		topicName:       shortLockTopic,
+		listener:        l,
+		publisher:       pub,
+		listenerOptions: []listener.Option{listener.WithMessageLockAutoRenewal(1 * time.Second)},
+		shouldSucceed:   true,
+	}, event)
+}
+
+func (suite *serviceBusSuite) TestPublishAndListenNotRenewingLock() {
```
serbrech, line 147: added a test that verifies behavior when lock is not renewed, expecting an error back on .Complete() calls

internal/test/test.go
```diff
@@ -48,6 +53,12 @@ func (suite *BaseSuite) SetupSuite() {
 	if err := godotenv.Load("../.env"); err != nil {
 		suite.T().Log(err)
 	}
+	if os.Getenv("TRACING") == "1" {
+		_, err := initTracing()
+		if err != nil {
+			suite.FailNow("failed to initialize tracing %s", err)
+		}
+	}
```
serbrech, line 61: sets up exporting to jaeger when env TRACING=1. see docker-compose env variable for export configuration variables

paulgmiller (approved):

Issue #17, opened 2021-01-22
Handle Message Expiration (TTL)


serbrech:
Messages that have been on the queue longer than the defined TTL reach expiration. They cannot be completed anymore.
When using RetryLater, they we could stay in memory without noticing that the message has expired.
This has to be handled in the listener, so that we stop delaying the in-memory message handling once the message expiration has been reached.

i.e : periodically check that `time.Now() < msg.SystemProperties.EnqueuedTime + message.TTL` and abandon the message when reached.

PR #21, opened 2021-01-31
feat: create a separate message context for each message

### LINKED ISSUE #19: Add per message context
Currently the context is shared across messages.
we need to create a new context for each being message handled to be able to cancel them independently on failures.

serbrech:
create a separate message context for each message in the processing loop.

This allows to attach a processing deadline equal to `enqueuedTime + TTL` and to avoid impacting other message processing. A side effect of this change is that the periodic renewer can rely on the context cancellation signal to stop renewing, and does not need to be explicitly stopped.

resolves #19 and #17

### DIFF
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -273,15 +273,16 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	// Create a handle class that has that function
 	listenerHandle := subReceiver.Listen(ctx, servicebus.HandlerFunc(
 		func(ctx context.Context, msg *servicebus.Message) error {
-			ctx, s := tab.StartSpan(ctx, "go-shuttle.Listener.HandlerFunc")
+			msgCtx, finished := messageContext(ctx, msg)
+			defer finished()
+			msgCtx, s := tab.StartSpan(msgCtx, "go-shuttle.Listener.HandlerFunc")
 			defer s.End()
 			if l.lockRenewalInterval != nil {
-				renewer := peeklock.RenewPeriodically(ctx, *l.lockRenewalInterval, sub, msg)
-				defer renewer.Stop()
+				peeklock.RenewPeriodically(msgCtx, *l.lockRenewalInterval, sub, msg)
 			}
 			currentHandler := handler
 			for !message.IsDone(currentHandler) {
-				currentHandler = currentHandler.Do(ctx, handler, msg)
+				currentHandler = currentHandler.Do(msgCtx, handler, msg)
 				// handle nil as a Completion!
 				if currentHandler == nil {
 					currentHandler = message.Complete()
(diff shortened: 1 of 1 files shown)


xhl873 (approved):
lgtm

listener/listener.go
```diff
@@ -298,6 +299,14 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	return listenerHandle.Err()
 }
 
+func messageContext(ctx context.Context, msg *servicebus.Message) (context.Context, func()) {
+	msgContext, finished := context.WithCancel(ctx)
+	if msg.SystemProperties.EnqueuedTime != nil && msg.TTL != nil {
+		msgContext, _ = context.WithDeadline(msgContext, msg.SystemProperties.EnqueuedTime.Add(*msg.TTL))
```
charliedmcb, line 305: When ```EnqueuedTime + TTL``` is encountered we will get to this when calling RetryLater:
https://github.com/Azure/go-shuttle/blob/742d9c2a5b8004ee72593938465142a46b662d61/message/retrylater.go#L35

I expect this is the behavior we want. I'm just mainly curious if the message is automatically removed from the topic at that time? I assume so?

charliedmcb: Also, do we want to add a similar context for ```EnqueuedTime + lock duration - 1``` following the TODO comment here:
https://github.com/Azure/go-shuttle/blob/742d9c2a5b8004ee72593938465142a46b662d61/message/retrylater.go#L32

charliedmcb: The part about completing after lock duration is strange because I remember it being a big issue/concern when the TODO comment was put there, but when I tried to repro it I don't think there was any issue.

Mainly just questioning if we need to handle the TODO comment or not.

charliedmcb (approved):
LGTM, mainly just a few curiosities

listener/listener.go
charliedmcb: I just realized that you were accessing EnqueuedTime and the TODO mentions dequeue time, so I expect there is still nothing to be done there as the comment states ```don't have access to dequeue time ```

listener/listener.go
```diff
@@ -273,15 +273,16 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
...
+			defer finished()
+			msgCtx, s := tab.StartSpan(msgCtx, "go-shuttle.Listener.HandlerFunc")
 			defer s.End()
 			if l.lockRenewalInterval != nil {
-				renewer := peeklock.RenewPeriodically(ctx, *l.lockRenewalInterval, sub, msg)
-				defer renewer.Stop()
+				peeklock.RenewPeriodically(msgCtx, *l.lockRenewalInterval, sub, msg)
 			}
 			currentHandler := handler
 			for !message.IsDone(currentHandler) {
```
charliedmcb, line 284: Not sure if this is a good idea or not, but do we possibly want to short circuit the operation here if msgCtx.Done() is true?

Any future calls to the handler would be pointless, correct?

listener/listener.go
charliedmcb: Still don't fully understand ctx, but from my understanding the context being closed will only impact the child go routine if they check ctx.Done, or used the context in some other way that reveals its closed. 

Curious if we need to add a check in the non-RetryLater handlers for ctx.Done(), as currently I don't see anything that would notice the context is closed.

serbrech:
Closing in favor of #23

PR #22, opened 2021-01-31
chore: split options out from main publisher/listener files

serbrech:
`listener.go` and `publisher.go` are becoming too large.
splitting the options out to keep them leaner


xhl873 (approved):

charliedmcb (approved):
Strict cut and paste?

LGTM :)

serbrech:
will be rebased once #23 merges

PR #23, opened 2021-02-01
feat: concurrent message handling

### LINKED ISSUE #8: Parallel message handling does not work
### Which version of the SDK was used?
github.com/Azure/go-shuttle v0.5.5
https://github.com/Azure/go-shuttle/tree/8da14c485b239d55fa6b75dd41704e7d3bd0e55f

### Which platform are you using? (ex: Windows, Linux, Debian)
Linux

### What problem was encountered?
How things are currently set up in go-shuttle I am only able to process one message from the service bus at a time, and don't get to move onto the next message until the first is completed. I have been unable to find any exposed feature that would allow me to handle concurrent messages.

From digging through the code base I found that there are a few locations in the dependency azure-service-bus-go that I believe are resulting in this behavior, with modification to go-shuttle I expect that either exposing a setting, a new AsyncListener or something similar could allow for concurrent handling of messages:
1. where handle message is called we get a blocking call until the handler completes the message:
https://github.com/Azure/azure-service-bus-go/blob/705d23958eb9c000582c03ba30413dc7e35eb25d/receiver.go#L235
2. PeakAndLock mode is the default, and not exposed to the user. From my understanding this means that the message will not be removed from the queue until it is completed meaning that we would not even be able to create additional Listeners to receive more messages, as we would still be blocked until the locked message was completed and removed (I expect this should not be changed, and most likely does not want to be exposed either):
https://github.com/Azure/azure-service-bus-go/blob/705d23958eb9c000582c03ba30413dc7e35eb25d/receiver.go#L108
3. The prefetchCount/linkCredit is set to 1 as default, and there is no setting exposed from go-shuttle to set it:
https://github.com/Azure/azure-service-bus-go/blob/705d23958eb9c000582c03ba30413dc7e35eb25d/receiver.go#L109

An additional note: as RetryLater simply pauses the process in memory, but doesn't remove the message from the queue we are unable to progress onto the next message until we actually complete it. I expect this in intended behavior, but wanted to make note.

### How can we reproduce the problem in the simplest way?
Create a Azure Service Bus namespace, and a topic
Setup a go-shuttle Listener for the topic
Enqueue multiple messages
Only one message will be handled at a time with no parallelism

### Have you found a mitigation/solution?
While working with a few others we have not found a solution yet, but have run some informative tests and have some ideas:
1. I modified code location 1. to ```go r.handleMessage(ctx, msg, handler)```, and code location 2. to ```mode:       ReceiveAndDeleteMode,```. When running a repro test with this I was successfully able to handle the messages in parallel. I expect modifying the first code location is probably not ok as I believe it possibly changes the current behavior, but might be an alright change. I believe it would need more knowledgeable voices in the mix and isn't even for this repo. However, I highly expect that modifying ```ReceiveAndDeleteMode``` is not an acceptable solution as it changes the default behavior. I believe even exposing the option to use it in go-shuttle is not an acceptable final solution as in my understanding in the event that the consuming service crashes the message would be lost.
2. Working with a few other people Xiahe found the following PR that gives an example of how to implement concurrency using the azure-service-bus-go lib: https://github.com/Azure/azure-service-bus-go/pull/117/files. Based of this PR, Paul Miller created a strawman PR in go-shuttle to emulate what the example PR showed: https://github.com/Azure/go-shuttle/pull/9. I ran a repro test using this PR and was able to confirm that it worked correctly; although, it was limited to handling the number of messages equal to Paul's concurrency value which is used to set the prefetchCount/linkCredit.

I believe the most likely solution will involve something similar to Paul's PR exposed as a new AsyncListener with an option to adjust what the concurrency value is set to, or modify the current Listener and add an exposed option to adjust what the concurrency is set to with the default staying at 1.

### LINKED ISSUE #17: Handle Message Expiration (TTL)

### LINKED ISSUE #19: Add per message context

serbrech:
resolves #8 
resolves #17 
resolves #19 
 
This creates a concurrent message handler that pumps the messages into a channel, and process them concurrently on the client side. 

- 🟢 passes all existing integration tests
- 🟢 write specific concurrency tests
- 🟢 expose prefetch settings (maps to amqp link credit)
- 🟢  add **unit** test where possible
- 🟢  expose max concurrency on the to balance the prefetch settings
- 🟢  add doc #24 

This PR starts from current main, there is a few other smaller PR that can get merged, and this one can then be rebased and build on top, in particular #21 (separate message context) which became necessary for concurent message handling.

~Also, I think I've hit this issue on some test runs when closing the listener : https://github.com/Azure/azure-uamqp-c/issues/228~

The above observation was fixed by ensuring we don't try to send message disposition after closing the connection.
see https://github.com/Azure/go-shuttle/pull/23/files#r568313406

## Handler Pipeline (middleware)

Just like http request/response middlewares, we have here Message middleware pipeline : 

`servicebus.Handler` is the interface that needs to be passed to the receiver.
with the following signature : 
```golang
Handler interface {
	Handle(context.Context, *Message) error
}
```

The handlers in the `handlers` I created all take as constructor parameter a `next servicebus.Handler` argument.
this allows each handler to do their work, and pass the message down to the next handler in the pipeline.

The pipeline I propose here is as follow : 
 
```golang
handlers.NewConcurrent(maxConcurrency,
	handlers.NewDeadlineContext(
		handlers.NewPeekLockRenewer(l.lockRenewalInterval, sub,
			handlers.NewShuttleAdapter(handler)),
	)))
```

`Concurrent`, `DeadlineContext`, `PeekLockRenewer`, `ShuttleAdapter`

- `Concurrent` is the entry point, it fans out the messages into the rest of the pipeline.
-  the first thing we do then, is create a message context to isolate the processing of each message. this context includes a deadline as documented in the code.
- now that we have a message context, we can start the PeekLock auto-renewal. we must have a message context to start it, as we rely on the message context cancellation to stop it.
- we then feed the message into the ShuttleAdapter, which is the loop we've had from before, waiting for the message handler to return a doneHandler, and does the sericebus.Message to our Message type translation

## Concurrent handler: 

I removed the buffering completely. AMQP is a streaming protocol on top of an always open TCP connection, so buffering doesn't really help anything, except hogging memory.
Instead I use a blocking token channel, similar to what we've used in the RP.

## DeadlineContext

Except from the Concurrent handler, each single message processing pipeline is synchronous, we know that when we return from a handler, all handlers below us are done doing their work.
the message context sits above the peeklock renewer, and above the message processing. so all network calls are done at this point. this allows us to cancel the context that we create for the message within the same handler, removing synchronization concerns.

Note: the deadline is calculated from the info on the message, which is subject to clock skew.

## PeekLock
we've discussed this one in a different PR, the logic is moved from the peeklock package into this middleware.

## ShuttleAdapter
Wraps the message and invokes the go-shuttle message state machine until we reach a done state


concurrent/handler.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func NewHandler(ctx context.Context, r peeklock.LockRenewer, renewalInterval *time.Duration, handler message.Handler) servicebus.Handler {
+	h := &Handler{
+		renewalInterval: renewalInterval,
+		lockRenewer:     r,
+		messageHandler:  handler,
+		messages:        make(chan *servicebus.Message),
+	}
+	go h.handleMessages(ctx)
```
serbrech, line 27: start looping on the message channel internally before returning the handler

concurrent/handler.go
```diff
@@ -0,0 +1,77 @@
...
+func (h *Handler) Handle(_ context.Context, msg *servicebus.Message) error {
+	h.messages <- msg
+	// servicebus handler errors interrupt the listener. A single message handling error should not stop the listener.
+	// always return nil here
+	return nil
+}
+
+func (h *Handler) handleMessages(ctx context.Context) {
+	for msg := range h.messages {
+		go h.handleMessage(messageContext(ctx, msg), msg)
```
serbrech, line 40: handle each message in a background goroutine

integration/pubsub_test.go
```diff
@@ -369,7 +372,7 @@ func (suite *serviceBusSuite) publishAndReceiveMessageWithRetryAfter(testConfig
 		if testConfig.shouldSucceed != isSuccessful {
 			suite.FailNow("Test did not succeed")
 		}
-	case <-time.After(5 * time.Second):
+	case <-time.After(10 * time.Second):
```
serbrech, line 375: can probably reduce this again

listener/listener.go
```diff
@@ -29,7 +27,7 @@ type Listener struct {
 	topicName           string
 	subscriptionName    string
 	maxDeliveryCount    int32
-	lockRenewalInterval *time.Duration
+	LockRenewalInterval *time.Duration
```
serbrech, line 30: if we can extract the lock renewal from the handler, this can go back to private.

message/retrylater.go
```diff
@@ -22,19 +22,20 @@ type retryLaterHandler struct {
 }
 
 func (r *retryLaterHandler) Do(ctx context.Context, orig Handler, message *servicebus.Message) Handler {
-	go func() {
```
serbrech, line 25: no need to have this in a go routine. each msg handler is already processed in a goroutine. let it block and make it simpler

integration/pubsub_test.go
```diff
@@ -547,35 +550,54 @@ func checkResultHandler(publishedMsg string, publishedMsgType string, ch chan<-
 	return message.HandleFunc(
 		func(ctx context.Context, msg *message.Message) message.Handler {
 			if publishedMsg != msg.Data() {
+				errHandler := message.Error(errors.New("published message and received message are different"))
```
serbrech, line 553: these changes are ensuring that we first run the Complete or Abandon handlers, before we push the result in the channel.
pushing the result in the channel will continue the processing in the test, which closes the connection, but Complete/Abandon send a frame on the amqp link. if we don't close the connection before, Complete/Abandon trigger unexpected errors

listener/listener.go
```diff
@@ -260,37 +275,23 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
...
-			for !message.IsDone(currentHandler) {
-				currentHandler = currentHandler.Do(ctx, handler, msg)
-				// handle nil as a Completion!
-				if currentHandler == nil {
-					currentHandler = message.Complete()
-				}
-			}
-			return nil
-		}))
+	listenerHandle := subReceiver.Listen(ctx, concurrent.NewHandler(ctx, sub, l.LockRenewalInterval, handler))
```
charliedmcb, line 292: I believe you need a .Handle() here

listener/listener.go
```diff
@@ -260,37 +275,23 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
...
+	// TODO: Add servicebus.SubscriptionWithPrefetchCount(prefetch count) option and expose the setting via the listener
 	sub, err := topic.NewSubscription(l.subscriptionEntity.Name)
 	if err != nil {
 		return fmt.Errorf("failed to create new subscription %s: %w", l.subscriptionEntity.Name, err)
 	}
-	subReceiver, err := sub.NewReceiver(ctx)
+
+	var receiverOpts []servicebus.ReceiverOption
+	if l.PrefetchCount != nil {
+		receiverOpts = append(receiverOpts, servicebus.ReceiverWithPrefetchCount(*l.PrefetchCount))
```
charliedmcb, line 286: I'm assuming the TODO on line 278 might be for this, but I wanted to let you know that this setting can be applied if the subscription has a prefetch count (Also, could be overridden if there was one as the following option is applied after; however, I don't see how the subscription would be non-nil at that point):
https://github.com/Azure/azure-service-bus-go/blob/12810d63bcdc551ef6446e18e08815cdbb2fc74d/subscription.go#L143

concurrent/handler.go
charliedmcb: I wanted to double check that is that pointless to set a client concurrency limit as it will be always limited by the link credit? I'm assuming we will only worry about the link credit?

charliedmcb (commented):
Mostly minor questions, and comments. Overall, looking great!

Note: I didn't review the test file. I can do that at a later time if you would like :)

message/retrylater.go
```diff
@@ -23,19 +23,17 @@ type retryLaterHandler struct {
 }
 
 func (r *retryLaterHandler) Do(ctx context.Context, orig Handler, message *servicebus.Message) Handler {
```
charliedmcb, line 25: Note: I believe this is a result of a previous PR that made the auto lock renewal, but the following should be updated:

The comment here should probably be updated:
https://github.com/Azure/go-shuttle/blob/bab7408d88474b667938609459f1bdb8cd6e28c3/message/retrylater.go#L14

Also, the README.md for go-shuttle:
https://github.com/Azure/go-shuttle#postponed-handling-of-message

charliedmcb: Simply should add some info about how the auto lock renewal affects this

listener/listener.go
serbrech: done

message/retrylater.go
serbrech: that assumes that you enable auto lock renewal :). but yeah, we know the behavior here. I'll update the comment

listener/listener.go
```diff
@@ -146,19 +146,34 @@ func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) ManagementOp
 		}
 		l.lockDuration = lock
 		if maxDelivery < 0 {
```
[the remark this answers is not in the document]
serbrech: will resolve this comment and create an issue instead

peeklock/renewer.go
```diff
@@ -15,6 +15,7 @@ type LockRenewer interface {
 type PeriodicLockRenewer struct {
 	lockRenewer LockRenewer
 	cancelFunc  func()
+	timer       *time.Timer
```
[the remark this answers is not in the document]
serbrech: gone

serbrech:
So, after running a lot of concurrent tests and understanding the prefetch/link credit functionality better, I paired with @devigned and simplified the concurrency.

David reminded me that the handlers are built as a middleware chain, and from there I found how to easily split up the concerns in separate small middlewares that handle 1 aspect of the message pipeline.
It becomes much clearer, easier to test, and easier to review.

The double whammy free lunch outcome from this, is that these handlers are pure servicebus SDK middleware, and as such, they can very easily be moved into the upstream project :) @devigned 👏.
only the last handler in the middleware chain is a go-shuttle adapter.

I outline the design in the PR description at the top

concurrent/handler.go
serbrech: You could say that AMQP is a push protocol (with flow control) so, as long as you have credits, the broker is pushing message on the TCP connection. They then sit on our side, "unsettled" and we still need to control the concurrency of our workers to process these before their peek lock expires. That is why the link credit (prefetch setting) needs to be balanced with the worker concurrency level to optimize throughput.

handlers/shuttleadapter.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/Azure/go-shuttle/message"
+	"github.com/devigned/tab"
+)
+
+var _ servicebus.Handler = (*ShuttleAdapter)(nil)
+
+// ShuttleAdapter wraps a go-shuttle message.Handler in a servicebus.Handler.
+// It allows the consumer to write go-shuttle handler func
+// that uses the go-shuttle semantic and message disposition API
+type ShuttleAdapter struct {
```
devigned, line 16: WDYT of the adapter or some other structure having a slice of middleware? Then when you actually want to materialize the final function you can build the handler from the stack of middleware like: https://github.com/devigned/pub/blob/8124492a2e15b32f2d0a55d9b6e7a71f33ed4478/pkg/partner/client.go#L557-L600.

The upside to this design is that you don't have to decorate next all over the place.

devigned (commented):
I really love how this turned out. It's super clean and decomposed the concerns into small chunks of easily understood and tested behaviors.

I left a comment for you to consider about how your middleware is stored and constructed.

listener/listener.go
```diff
@@ -260,37 +286,32 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
...
-			for !message.IsDone(currentHandler) {
-				currentHandler = currentHandler.Do(ctx, handler, msg)
-				// handle nil as a Completion!
-				if currentHandler == nil {
-					currentHandler = message.Complete()
-				}
-			}
-			return nil
-		}))
+	listenerHandle := subReceiver.Listen(ctx,
```
charliedmcb, line 307: Seems good as is. 

However, I am curious what making the handlers public is meant to achieve? Are we thinking the user of this lib, might construct their own subReceiver and their own set of handlers? Seems like a viable; although, unlikely option. Simply curious :)

charliedmcb (commented):
Reviewed updated since last review minus testing files

Mostly nits and curiosity

listener/listener.go
serbrech: ah, so azure servicebus exposes it on both Subscription and NewReceiver.
setting it on the Subscription object ends up passing it down to the receiver. I pass it directly to the receiver: 

```golang
// NewReceiver will create a new Receiver for receiving messages off of the queue
func (s *Subscription) NewReceiver(ctx context.Context, opts ...ReceiverOption) (*Receiver, error) {
	ctx, span := s.startSpanFromContext(ctx, "sb.Subscription.NewReceiver")
	defer span.End()

	opts = append(opts, ReceiverWithReceiveMode(s.receiveMode))

	if s.prefetchCount != nil {
		opts = append(opts, ReceiverWithPrefetchCount(*s.prefetchCount))
	}

	return s.namespace.NewReceiver(ctx, s.Topic.Name+"/Subscriptions/"+s.Name, opts...)
}
```

handlers/shuttleadapter.go
serbrech: Took the discussion offline.
Basically there is 2 approaches to middleware pipeline: 
- having an explicit pipeline that manage the flow from a list of middleware
- or the one I did, which chains the middlewares explicitly, where they call the next one each time.

I chose to chain them, because I don't think the middleware complexity is going to grow more than what it is now.
If it does, we can reconsider, but I think this works well, so I'd like to keep it this way.

handlers/messagecontext.go
```diff
@@ -0,0 +1,50 @@
...
+// DeadlineContext is a servicebus middleware handler that creates a message context from the incoming context to isolate
+// each message handling sub processes. It adds a deadline equal to `EnqueuedTime + TTL` to the message context if the message TTL is defined.
+// Otherwise no deadline is applied on the message context.
+// The context is canceled once the `next` middleware returns. this implies that no middleware above DeadlineContext can use this context.
+type DeadlineContext struct {
+	next servicebus.Handler
+}
+
+func (mc *DeadlineContext) Handle(ctx context.Context, msg *servicebus.Message) error {
+	if mc.next == nil {
```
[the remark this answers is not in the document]
serbrech: you could still create the handler via the public struct. but I'll add it there too yeah.

handlers/messagecontext.go
```diff
@@ -0,0 +1,50 @@
...
+		return NextHandlerNilError
+	}
+	msgCtx, cancel := setupDeadline(ctx, msg)
+	// NOTE: this assumes that no handler above the DeadlineContext handler needs this context.
```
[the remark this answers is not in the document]
serbrech: yes but these handler could be re-organized in a different way by a user who uses ServiceBus SDK directly, or builds his own listener. they are public and reusable on their own.

listener/listener.go
serbrech: yes. as made clear in the description, these handler are not at all tied to go-shuttle. they can be used with the raw sdk, and they will likely move there in the future.

handlers/shuttleadapter.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func NewShuttleAdapter(next message.Handler) *ShuttleAdapter {
+	return &ShuttleAdapter{next: next}
+}
+
+func (c *ShuttleAdapter) Handle(ctx context.Context, msg *servicebus.Message) error {
+	ctx, s := tab.StartSpan(ctx, "go-shuttle.Listener.HandlerFunc")
+	defer s.End()
+	if c.next == nil {
```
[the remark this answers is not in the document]
serbrech: you don't want to return an error from a constructor, I can't prevent the user from passing nil. and I don't think you should panic from a ctor either. So that's the best place to validate I think.

handlers/lockrenewer.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (plr *PeekLockRenewer) startPeriodicRenewal(ctx context.Context, message *servicebus.Message) {
+	_, span := tracing.StartSpanFromMessageAndContext(ctx, "go-shuttle.peeklock.startPeriodicRenewal", message)
+	defer span.End()
+	for alive := true; alive; {
+		select {
+		case <-time.After(*plr.renewalInterval):
+			span.Logger().Debug("Renewing message lock")
+			err := plr.lockRenewer.RenewLocks(ctx, message)
+			if err != nil {
```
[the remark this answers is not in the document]
serbrech: yeah it could be, but RenewLocks uses the servicenus internal rpc client, which has retries built-in.
either way, it's better to bail and let the queueing system retry the message, which is what would happen eventually when RenewLock fails.

handlers/lockrenewer.go
```diff
@@ -0,0 +1,64 @@
...
+	next            servicebus.Handler
+	lockRenewer     LockRenewer
+	renewalInterval *time.Duration
+}
+
+func (plr *PeekLockRenewer) Handle(ctx context.Context, msg *servicebus.Message) error {
+	if plr.next == nil {
+		return NextHandlerNilError
+	}
+	if plr.renewalInterval != nil {
```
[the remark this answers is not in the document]
serbrech: no, the handler is always in the pipeline. when you don't set auto-renewal, the value is nil, and so the middleware becomes pass-through (no-op), which is fine. easier than to build the middleware pipeline depending on the settings, and pretty much no impact.

paulgmiller:
Thanks for the great write up in the PR description here!
"I removed the buffering completely. " By that I assume you mean you're not doing prefetch?  Maybe I'll understand as I read.

handlers/lockrenewer.go
```diff
@@ -0,0 +1,64 @@
...
+
+// LockRenewer abstracts the servicebus subscription client where this functionality lives
+type LockRenewer interface {
+	RenewLocks(ctx context.Context, messages ...*servicebus.Message) error
+}
+
+// PeekLockRenewer starts a background goroutine that renews the message lock at the given interval until Stop() is called
+// or until the passed in context is canceled.
+// it is a pass through handler if the renewalInterval is nil
+type PeekLockRenewer struct {
```
paulgmiller, line 22: Did you never complete the peeklock package PR or just wiping it in another PR. I assum not much has changed here.

handlers/lockrenewer.go
```diff
@@ -0,0 +1,64 @@
...
+	defer span.End()
+	for alive := true; alive; {
+		select {
+		case <-time.After(*plr.renewalInterval):
+			span.Logger().Debug("Renewing message lock")
+			err := plr.lockRenewer.RenewLocks(ctx, message)
+			if err != nil {
+				// I don't think this is a problem. the context is canceled when the message processing is over.
+				// this can happen if we already entered the interval case when the message is completing.
+				span.Logger().Info("failed to renew the peek lock", tab.StringAttribute("reason", err.Error()))
```
paulgmiller, line 57: warn or error?

paulgmiller: still feel like this should cancel the context if we repeatedly can't renew but I think we already talked about that in the other peek lock pr so not worth rehashing here.

listener/listener.go
```diff
@@ -146,13 +147,38 @@ func WithSubscriptionDetails(lock time.Duration, maxDelivery int32) ManagementOp
...
+			return fmt.Errorf("max deliveries must be positive")
 		}
 		l.maxDeliveryCount = maxDelivery
 		return nil
 	}
 }
 
+// WithPrefetchCount the receiver to quietly acquires more messages, up to the PrefetchCount limit. A single Receive call to the ServiceBus api
+// therefore acquires several messages for immediate consumption that is returned as soon as available.
+// Please be aware of the consequences : https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-prefetch#if-it-is-faster-why-is-prefetch-not-the-default-option
```
paulgmiller, line 159: should we only set this when concurrency is on /== to prefetch?

listener/listener.go
paulgmiller, line 307: this is adorably clean now :) Good job.

paulgmiller (approved):
I had to do some pretty fast skimming since my work day is cut in half but excite about the direction. Hopefully Xiahe and Charlie got you some feedback.

message/retrylater.go
```diff
@@ -23,19 +23,17 @@ type retryLaterHandler struct {
...
-		}
-	}()
+	select {
+	// TODO this can go past lock duration pretty easily
+	// Ideally we'd also timeout at dequeue time + lockdurtaion - 1 second but don't have access to dequeue time
+	// Maybe we can use context.WithTimeout when we recieve the message?
+	case <-ctx.Done():
+		span.Logger().Error(errors.Wrap(ctx.Err(), "Retry context expired"))
+	case <-time.After(r.retryAfter):
+		Abandon().Do(ctx, orig, message)
```
paulgmiller, line 36: should we stop abandoning now that we have peek lock?

handlers/lockrenewer.go
paulgmiller: Looks like it's there. Git rm it or some reason to wait?

handlers/lockrenewer.go
serbrech: you can't cancel the message context. you don't have the cancel func, and
it would be really weird to pass that around IMO. you still have the same final behavior. the message is abandoned.
if you don't renew, then you still leave a chance to the handler to complete before the lock expires. but most likely, you're going to fail the Complete call too. unless it's a network fluke.
One way to potentially handle this is to put the cancel func on the context as a value? (is that possible?)
still, would be bending over backwards. I'd say we'll see if that's actually needed after some mileage.

serbrech: you don't have warn with Spans. it's error or Info. :)

message/retrylater.go
serbrech: mmh, I don't think we should.
you retry after a given time.
you can say after 15 minutes... even if your lock is only 2 min, that will work thanks to the auto-renewal.

plus, it makes the handler cleaner.
I can now remove the unused orginalHandler passed to every handler :)

Issue #25, opened 2021-02-04
Handle errors on listener/publisher creation


serbrech:
In real world applications, there is more than 1 replica of the service listening or publishing messages.

Currently, when concurrent PUT request are sent to Azure to create the same topic, the first request succeeds, but the subsequent ones will return 409 Conflict http response until the topic is successfully created.

We also hit this problem when starting all integration tests in parallel, which forced us to select which ones can run concurrently.

The library should **reasonably** retry these requests to allow multiple replicas of a service to start simultaneously

Suggestion: add simple retries on management api requests

PR #28, opened 2021-02-04
feat: add a simple retry on ensureTopic

serbrech:
Simple retry on Put topic to avoid 409 when concurrent publisher start up.

### DIFF
--- a/publisher/publisher.go
+++ b/publisher/publisher.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	common "github.com/Azure/azure-amqp-common-go/v3"
 	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-shuttle/internal/aad"
@@ -147,6 +148,9 @@ func SetCorrelationID(correlationID string) Option {
 
 // New creates a new service bus publisher
 func New(topicName string, opts ...ManagementOption) (*Publisher, error) {
+	ctx := context.Background()
+	ctx, s := tab.StartSpan(ctx, "go-shuttle.publisher.New")
+	defer s.End()
 	ns, err := servicebus.NewNamespace()
 	if err != nil {
 		return nil, err
@@ -158,7 +162,8 @@ func New(topicName string, opts ...ManagementOption) (*Publisher, error) {
 			return nil, err
 		}
 	}
-	topicEntity, err := ensureTopic(context.Background(), topicName, publisher.namespace, publisher.topicManagementOptions...)
+
+	topicEntity, err := ensureTopic(ctx, topicName, publisher.namespace, publisher.topicManagementOptions...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get topic: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


serbrech:
issue #25

publisher/publisher.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	common "github.com/Azure/azure-amqp-common-go/v3"
```
serbrech, line 10: hmm, is that ok? I'm really just using a dead simple retry... autorest has one too. thoughts?

publisher/publisher.go
```diff
@@ -147,6 +148,9 @@ func SetCorrelationID(correlationID string) Option {
 
 // New creates a new service bus publisher
 func New(topicName string, opts ...ManagementOption) (*Publisher, error) {
+	ctx := context.Background()
```
serbrech, line 151: really this should come from the caller, but that changes the api.
so keeping it for a separate PR.

serbrech: created #29

publisher/publisher.go
```diff
@@ -208,10 +213,23 @@ func (p *Publisher) Publish(ctx context.Context, msg interface{}, opts ...Option
...
+			return te, nil
+		}
+		te, err = tm.Put(ctx, name, opts...)
+		if err != nil {
+			// let all errors be retryable for now. application only hit this once on topic creation.
+			return nil, common.Retryable(err.Error())
+		}
 		return te, nil
 	}
+	entity, err := common.Retry(5, 1*time.Second, ensure)
```
serbrech, line 228: max 5 sec toll on service start-up seems reasonable.
test shows that it passes on 2nd attempt, so could reduce this further.

publisher/publisher.go
paulgmiller: You're not getting much from it you wouldn't get from avast. Your biggest risk is that they change how they decide when to retry but Presuabley they'd change common.Retryable at that point.

publisher/publisher.go
```diff
@@ -208,10 +213,23 @@ func (p *Publisher) Publish(ctx context.Context, msg interface{}, opts ...Option
...
-	te, err := tm.Get(ctx, name)
-	if err == nil {
+	ensure := func() (interface{}, error) {
+		te, err := tm.Get(ctx, name)
+		if err == nil {
+			return te, nil
+		}
+		te, err = tm.Put(ctx, name, opts...)
+		if err != nil {
+			// let all errors be retryable for now. application only hit this once on topic creation.
```
paulgmiller, line 223: Even 429's? Vague preferenc to just retry >500 and 409 but you retry is slow enough it's not a big deal.

publisher/publisher.go
paulgmiller: You could explictly re-get on 409 and skip the retry all together assuming that callers of ensureTopic crash the program and it will just get restarted by k8s other orchestrator.

publisher/publisher.go
paulgmiller: Question is how easy is the status code to pull out of the err. If it's string matching then agreee retrying all might be better.

paulgmiller (approved):
Some qualms about tiny retries of all errors but I think this is mostly harmless.

publisher/publisher.go
serbrech: yes, but crashing the program is not working out for e2e tests, cause every runs was causeing the replicas to crash at least once.

publisher/publisher.go
serbrech: easy, yes, easy.
reliable... well, I depend on the implementation of the azure-serbice-bus-go.
granted, we own it, and any change in that package has to be accounted for in go-shuttle, so probably very low risk.
just extra complexity that I don't think will bring value in practice.

PR #35, opened 2021-02-17
Add Explicit Subscription Flags for AZ CLI commands

imiller31:
In the case where the `default az-cli subscription id` != `$AZURE_SUBSCRIPTION_ID`, the current `test-setup` and `integration` scripts will fail.

`MANAGED_IDENTITY_RESOURCE_ID` in `test-setup.sh` will be constructed incorrectly due to the mismatch.

This PR fix just propagates the `AZURE_SUBSCRIPTION_ID` to all `az-cli` calls as an explicit `--subscription` flag.


serbrech:
thanks!

PR #38, opened 2021-03-15
Publisher with context

### LINKED ISSUE #29: publisher.New should take in a context

coip:
option that might resolve #29, scrutiny is appreciated.

half-baked backwards compat for API change on the top-level `NewPublisher()`, but does incur breaking change on `publisher/publisher.go`

commits are: pkg/lib update, test suite update, docs update.

lmk your thoughts.
would like to use this pkg instead of the vanilla Azure/azure-service-bus-go module.


serbrech:
Hi @coip thanks so much for rolling up your sleeves and contributing here!
As you have guessed, this is quite a young project. 
I created it to facilitate my team's integrations. so, we have the luxury of not being too worried about breaking changes (until you showed up 😄 )

I think we should just drop the top level `NewPublisher` funcs. they just haven't been cleaned up after the refactoring.
I'll take a closer look later today, but looks fine at first pass.

coip:
I figured that might be a better route for concise/intuitive structure. (esp given the v0 _breaking changes almost guaranteed_ angle)

Didnt want to make such a design decision given issue, but happy to reduce cruft and doubly happy to be an early contributor on a project such as this.

serbrech (approved):

PR #40, opened 2021-04-12
Separate lock duration and max count management option

xhl873:
currently lockduration and max delivery count need to be set use the same function, separating them.


paulgmiller (approved):
THanks for addressing Paul's prior laziness :)

PR #45, opened 2021-05-03
add publisher recovery on network error

serbrech:
When the underlying tcp connection is lost, the next publish call fails with a network error. The connection will never be re-established by AMQP, leaving the publisher in a truly broken state. the only available recourse for the application code is to re-create the publisher.

This PR adds recovery code in the `Send` func to rebuild the underlying topic and retry to send the message when we meet a network error that is not temporary, or when it is recognized as a timeout (happens when tcp has been closed).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.14
 require (
 	github.com/Azure/azure-amqp-common-go/v3 v3.1.0
 	github.com/Azure/azure-service-bus-go v0.10.12-0.20210225175612-bb0d9d0bf8ba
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -5,7 +5,6 @@ go 1.14
 require (
 	github.com/Azure/azure-amqp-common-go/v3 v3.1.0
 	github.com/Azure/azure-service-bus-go v0.10.12-0.20210225175612-bb0d9d0bf8ba
-	github.com/Azure/go-amqp v0.13.5
```
serbrech, line 8: go mod tidy removed the require.

PR #47, opened 2021-05-24
Abstracting Publisher

imiller31:
This change abstracts Publisher into an interface to support future ServiceBus `Queue` support.


imiller31:
@serbrech thoughts on this?

PR #48, opened 2021-05-26
Queue Publisher Implementation

imiller31:
Implementation and refactor to support ServiceBus Queues.

This PR focuses on establishing the pattern for creating a Queue Publisher -- the patterns decided on will follow for the Listener implementation/refactor.

### DIFF
--- a/integration/concurrentcreate_topic_test.go
+++ b/integration/concurrentcreate_topic_test.go
@@ -6,16 +6,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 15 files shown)


serbrech:
Looks fine to me. We'll refactor duplicated code in subsequent PRs. That will let us progress on the consumer side.

Probably would like to share the options if we manage to get a common api that makes sense

serbrech:
Haven't had the time to fix the integration test action. I want it to run before we merge

imiller31:
@serbrech let me know if you need help fixing the test actions, don't have permissions to take a look currently

PR #53, opened 2021-07-05
Instrument library with prometheus metrics

serbrech:
Allow to register a set of metrics from the library in a user provided prometheus registerer.

```golang
import (
    shuttle "github.com/Azure/go-shuttle/prometheus"
    prom "github.com/prometheus/client_golang/prometheus"
)

func init() {
    shuttle.Register(prom.DefaultRegisterer)
} 
```


bcho (approved):
LGTM

PR #56, opened 2021-07-14
fix deadline context leak

serbrech:
When an incoming has a TTL set, the message context handler was returning a deadline context, but with the cancel func from the original cancel context.
this means that the wrong cancel was being called and that the LockRenewer was running indefinitely in the background, creating sessions on the link that never got closed. it eventually reaches "max session" and stops working.

This returns the correct context, which will stop the renewer as soon as a message is done being handled.

### DIFF
--- a/handlers/messagecontext_test.go
+++ b/handlers/messagecontext_test.go
@@ -1,0 +1,42 @@
+package handlers_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	servicebus "github.com/Azure/azure-service-bus-go"
+	"github.com/Azure/go-shuttle/handlers"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_CancelContextWhenDoneWithMessage(t *testing.T) {
+	next := &FakeHandler{}
+	handler := handlers.NewDeadlineContext(next)
+	ctx := context.Background()
+	ttlMessage := &servicebus.Message{}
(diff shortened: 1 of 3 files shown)


handlers/messagecontext_test.go
```diff
@@ -0,0 +1,42 @@
...
+	servicebus "github.com/Azure/azure-service-bus-go"
+	"github.com/Azure/go-shuttle/handlers"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_CancelContextWhenDoneWithMessage(t *testing.T) {
+	next := &FakeHandler{}
+	handler := handlers.NewDeadlineContext(next)
+	ctx := context.Background()
+	ttlMessage := &servicebus.Message{}
```
charliedmcb, line 17: nit: `emptyMessage` instead of `ttlMessage`

handlers/messagecontext_test.go
```diff
@@ -0,0 +1,42 @@
...
+	err := handler.Handle(ctx, ttlMessage)
+	assert.Nil(t, err)
+	assert.Equal(t, context.Canceled, next.Ctx.Err())
+}
+
+func Test_CancelContextWhenDeadlineExceeded(t *testing.T) {
```
charliedmcb, line 23: This actually tests what was already happening. If we want to test the bug it would have to be cancel is called when deadline is not exceeded.

charliedmcb (approved):

PR #61, opened 2021-08-03
Queue Listener + Project Structure Flattening

imiller31:
refactor of the package structure for easier consumption -- flattening to allow for more natural `queue.NewListener()` type pattern.


common/baseinterfaces/publisher.go
```diff
@@ -0,0 +1,38 @@
+package baseinterfaces
+
+import servicebus "github.com/Azure/azure-service-bus-go"
+
+var _ BasePublisher = &PublisherSettings{}
+
+type BasePublisher interface {
```
serbrech, line 7: that will work.
just not fond of the naming. it's not go-like

what agbout `common/types.go`
and the type can just be Publisher.

so when it's used it's `common.Publisher` embedded in `queue.Publisher` or `topic.Publisher`

imiller31: agreed, that's a cleaner convention, changed

publisher/alias.go
```diff
@@ -1,73 +0,0 @@
-package publisher
```
serbrech, line 1: shouldn't that stay?

serbrech (approved):

PR #64, opened 2021-08-06
add delete subscription function

xhl873:
the current change added a function to clean up subscription that we don't need anymore. we can all it with old subscription names. There might be other options to do that, but I feel like they might be risky, 
1. in listener, list all subscriptions, delete other subscriptions that not currently in use. this will only allow one subscription, so don't think it's the right thing to do. 
2. use servicebus option SubscriptionWithAutoDeleteOnIdle (assuming we can set it on existing subscription), we can set a time window, the subscription will be deleted after the time window. The concern is if our worker is offline for a while, the subscription might be deleted, and when the worker comes back online, a new subscription will be created, but we may lose the messages in between.

### DIFF
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -164,6 +164,32 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	return listenerHandle.Err()
 }
 
+// DeleteSubscriptions deletes the giving subscriptions
+func (l *Listener) DeleteSubscriptions(ctx context.Context, topicName, subscriptionName []string) error {
+	if l.topicEntity == nil {
+		return fmt.Errorf("entity of topic is nil")
+	}
+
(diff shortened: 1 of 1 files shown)


listener/listener.go
```diff
@@ -164,6 +164,32 @@ func (l *Listener) Listen(ctx context.Context, handler message.Handler, topicNam
 	return listenerHandle.Err()
 }
 
+// DeleteSubscriptions deletes the giving subscriptions
```
serbrech, line 167: I don't think that it is a good API to provide "Delete an arbitrary subscription name" from a listener to a different subscription name. Listener is to listen.
The api is available at a lower level, on the "subscription manager", via the namespace.
that should be enough

xhl873: okay, I am also okay to use the api directly, created a separate pr for it. closing this one.

PR #69, opened 2021-09-08
bump sb dependencies

serbrech:
pulls latest improvement in RPC calls to use a single link.
fixes failure in renewlocks background go-routine


paulgmiller (approved):
keeping up with the jones is good.

PR #75, opened 2021-11-05
Adding DeadLetter functionality to a message

imiller31:
Allows for a given message to be put onto the deadletter queue


serbrech (approved):
Probably want to add a metric as well here. But can be in a different PR

PR #76, opened 2022-01-18
Adding forward DLQ to management option

imiller31:
adding ability to set a custom queue for DLQ forwarding

### DIFF
--- a/integration/publisher_queue_test.go
+++ b/integration/publisher_queue_test.go
@@ -44,6 +44,24 @@ func (suite *serviceBusQueueSuite) TestCreatePublisherUsingExistingQueue() {
 	}
 }
 
+// TestCreatePublisherWithDeadLetterForwardUsingNewQueue tests the creation of a publisher for a new queue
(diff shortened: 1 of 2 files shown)


serbrech:
LGTM
the integration test failure is ACI related when using MSI.
the other auth mechanism worked fine, so I'd say merge it

integration/publisher_queue_test.go
```diff
@@ -44,6 +44,24 @@ func (suite *serviceBusQueueSuite) TestCreatePublisherUsingExistingQueue() {
 	}
 }
 
+// TestCreatePublisherWithDeadLetterForwardUsingNewQueue tests the creation of a publisher for a new queue
+func (suite *serviceBusQueueSuite) TestCreatePublisherWithDeadLetterForwardUsingNewQueue() {
+	suite.T().Parallel()
```
serbrech, line 49: cache the suite.T()
otherwise you might run require calls on whatever the current T is in the suite...

PR #87, opened 2022-12-07
initial tracing setup

erinborders:
Adding distributed tracing handler


v2/processor.go
```diff
@@ -118,6 +120,43 @@ func NewPanicHandler(handler HandlerFunc) HandlerFunc {
...
+	return func(ctx context.Context, settler MessageSettler, message *azservicebus.ReceivedMessage) {
+		var err error
+		if message != nil && message.ApplicationProperties != nil {
+			ctx, err = ExtractMessageContext(message)
+			if err != nil {
+				// do something
+			}
+		}
+
+		ctx, span := tab.StartSpan(ctx, "")
```
serbrech, line 134: need to define a naming prefix for go-shuttle spans and keep it consistent.

v2/processor.go
```diff
@@ -118,6 +120,43 @@ func NewPanicHandler(handler HandlerFunc) HandlerFunc {
...
+				tab.StringAttribute("message.id", message.MessageID),
+				tab.StringAttribute("message.correlationId", *message.CorrelationID))
+			if message.ScheduledEnqueueTime != nil {
+				span.AddAttributes(tab.StringAttribute("message.scheduledEnqueuedTime", message.ScheduledEnqueueTime.String()))
+			}
+			if message.TimeToLive != nil {
+				span.AddAttributes(tab.StringAttribute("message.ttl", message.TimeToLive.String()))
+			}
+		} else {
+			span.Logger().Error(fmt.Errorf("message is nil"))
```
serbrech, line 146: Maybe Warn, instead of Error.

erinborders: Okay, I didn't see a warn method for the span logger so I used info and just included warning in the message

v2/processor.go
erinborders: Ok made the change - I modeled the naming prefix off what I saw in the original go-shuttle, what do you think?

v2/processor.go
```diff
@@ -169,3 +199,20 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
...
 	}
 }
+
+type MessageWrapper struct {
+	Message *azservicebus.ReceivedMessage
+}
+
+// Set implements tab.Carrier interface
+func (mw *MessageWrapper) Set(key string, value interface{}) {
+	if mw.Message.ApplicationProperties == nil {
```
serbrech, line 209: check if Message is nil too. we don't want this code to cause panics in user's message handling

v2/processor.go
```diff
@@ -118,6 +119,35 @@ func NewPanicHandler(handler HandlerFunc) HandlerFunc {
 	}
 }
 
+// NewTracingHandler extracts the context from the message Application property if available, or from the existing
```
serbrech, line 122: Could you move the tracing handler to v2/tracing/tab?

serbrech:
this looks good :)
I pushed a few small changes that caused minor conflicts though

v2/processor.go
```diff
@@ -184,6 +183,28 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
 	}
 }
 
+type MessageWrapper struct {
```
serbrech, line 186: these should move to the tracing package and be private.

serbrech (approved):

PR #89, opened 2022-12-13
implement log and metrics integration support

serbrech:
Log integration allows to configure goshuttle package to use a `GetLogger(context.Context)` func to retrieve a logger.
this allows a per message logger to be retrieved, and used by our library if configured.

Metrics integration is a direct port from v1 to v2.

### DIFF
--- a/v2/metrics/registry.go
+++ b/v2/metrics/registry.go
@@ -58,11 +58,12 @@ func getMessageTypeLabel(msg *azservicebus.ReceivedMessage) prom.Labels {
 }
 
 func (m *Registry) Init(reg prom.Registerer) {
-	reg.MustRegister(m.MessageHandledCount,
+	reg.MustRegister(
+		m.MessageReceivedCount,
+		m.MessageHandledCount,
 		m.MessageLockRenewedCount,
 		m.MessageDeadlineReachedCount,
-		m.ConcurrentMessageCount,
-		m.ConnectionRecovery)
+		m.ConcurrentMessageCount)
 }
 
 type Registry struct {
@@ -71,7 +72,6 @@ type Registry struct {
 	MessageLockRenewedCount     *prom.CounterVec
 	MessageDeadlineReachedCount *prom.CounterVec
 	ConcurrentMessageCount      *prom.GaugeVec
-	ConnectionRecovery          *prom.CounterVec
 }
 
 type Recorder interface {
--- a/v2/metrics/registry.go
+++ b/v2/metrics/registry.go
@@ -1,0 +1,166 @@
+package metrics
+
+import (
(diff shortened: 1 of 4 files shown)


v2/metrics/registry.go
```diff
@@ -0,0 +1,179 @@
...
+func (m *Registry) IncMessageLockRenewedFailure(msg *azservicebus.ReceivedMessage) {
+	labels := getMessageTypeLabel(msg)
+	labels[successLabel] = "false"
+	m.MessageLockRenewedCount.With(labels).Inc()
+}
+
+// IncMessageHandled increase the message Handled
+func (m *Registry) IncMessageHandled(msg *azservicebus.ReceivedMessage) {
+	labels := getMessageTypeLabel(msg)
+	labels[deliveryCountLabel] = fmt.Sprintf("%d", msg.DeliveryCount)
```
jaiveerk, line 108: super nit: can we use strconv.formatInt here? I think it's more readable (plus a bit faster)

v2/metrics/registry.go
```diff
@@ -0,0 +1,179 @@
...
+	m.MessageDeadlineReachedCount.With(labels).Inc()
+}
+
+// IncMessageReceived increases the message received counter
+func (m *Registry) IncMessageReceived(count float64) {
+	m.MessageReceivedCount.With(map[string]string{}).Add(count)
+}
+
+// Informer allows to inspect metrics value stored in the registry at runtime
+type Informer struct {
```
jaiveerk, line 134: more of a question than feedback but - we shouldn't be using the informer outside of testing purposes according to Prometheus' conventions, right? But I'm guessing that we're doing this because at some point we'll need to know if we experience failure trying to renew a lock on a message?

v2/processor.go
```diff
@@ -151,41 +157,60 @@ type peekLockRenewer struct {
...
 
-// dumb log until we integrate logging
-func log(a ...any) {
+type Logger interface {
+	Info(s string)
+	Warn(s string)
+	Error(s string)
+}
+
+var getLogger = func(_ context.Context) Logger { return &printLogger{} }
```
jaiveerk, line 196: these are just placeholders right?

jaiveerk (approved):

v2/metrics/registry.go
serbrech: we are.
for example, it allows you to use the metrics data in the Health endpoint of your service. we do that in Fleet and in AKS. why should you not do that?

v2/processor.go
serbrech: no, this is the default print logger if you don't set a logger but enable logging.

PR #91, opened 2022-12-15
Initial Sender Implementation

jaiveerk:


v2/sender/sender.go
```diff
@@ -0,0 +1,83 @@
...
+	"reflect"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
+)
+
+// MessageBody is a type to represent that an input message body can be of any type
+type MessageBody any
+
+// SendHandlerFunction is a function that is used to modify the ServiceBus message before sending
+type SendHandlerFunction func(ctx context.Context, msg *azservicebus.Message)
```
serbrech, line 15: `SendHandlerFunc`

v2/sender/sender.go
```diff
@@ -0,0 +1,83 @@
...
+
+// CustomSender is a wrapper around the ServiceBus sender that allows for users to introduce middleware to modify the ServiceBus message before it's sent
+type CustomSender interface {
+	SendMessage(ctx context.Context, mb MessageBody, options *azservicebus.SendMessageOptions, middleware SendHandlerFunction) error
+}
+
+// Sender is an interface that allows users to substitute their own senders for the default ServiceBus sender when making a CustomSender
```
serbrech, line 22: // Sender is satisfied by *azservicebus.Sender

serbrech: maybe call this SBSender to not conflict with goshuttle.Sender?

v2/sender/sender.go
```diff
@@ -0,0 +1,83 @@
...
+// DefaultSender contains a Sender used to send the message to the ServiceBus queue and a Marshaller used to marshal any struct into a ServiceBus message
+type DefaultSender struct {
+	sender     Sender
+	marshaller Marshaller
+}
+
+var _ CustomSender = &DefaultSender{}
+
+// NewDefaultSender takes in a Sender and a Marshaller to create a new object that can send messages to the ServiceBus queue
+func NewDefaultSender(sender Sender, marshaller Marshaller) *DefaultSender {
```
serbrech, line 36: NewSender

serbrech: take options, not Marshaller directly

v2/sender/sender.go
```diff
@@ -0,0 +1,83 @@
...
+	return &DefaultSender{sender: sender, marshaller: marshaller}
+}
+
+// SendMessage marshals the input struct, runs middleware to modify the returned ServiceBus message, and uses the Sender to send the message to the ServiceBus queue
+func (d *DefaultSender) SendMessage(ctx context.Context, mb MessageBody, options *azservicebus.SendMessageOptions, middleware SendHandlerFunction) error {
```
serbrech, line 41: since we apply a middleware func on the SendMessage, let's roll the azserviecus.SendMessageOptions as a SendHandlerFunc to clean up the function signature

serbrech, line 41: rename `middleware SendHandlerFunction`
to options SendOption

v2/sender/sender.go
```diff
@@ -0,0 +1,83 @@
...
+		return fmt.Errorf("failed to send message: %s", err)
+	}
+
+	return nil
+}
+
+// SetTypeHandler sets the ServiceBus message's type to the original struct's type
+func SetTypeHandler(mb MessageBody, handler SendHandlerFunction) SendHandlerFunction {
+	return func(ctx context.Context, msg *azservicebus.Message) {
+		var msgType string
```
serbrech, line 64: should we implement this as a func on the MessageBody type?

`func (mb MessageBody) GetMessageType() string`

v2/sender/marshaller.go
```diff
@@ -0,0 +1,80 @@
+package sender
```
serbrech, line 1: can we try to keep this at the root of the package?
if there are things we don't want to expose, we can move them in an `internal` package in a refactoring

serbrech (changes requested):
few comments. direction is about right 👍

v2/sender/sender.go
jaiveerk: Unless I'm missing something, we can't make it a method since the MessageBody type is an interface, but I can still pull this out to a `GetMessageType(mb MessageBody) string` function?

v2/sender/sender.go
jaiveerk: I'm not sure I understand -
do you mean to take a SendOption (aka a handler/middleware)? If we do that, we don't have a guaranteed step in Send that converts the message body to a SB message, and so we can't feed the message/messagebody into the middleware chain.
Or, do you mean to abstract the Marshaller type into a separate "option" interface that has a method to marshal the messagebody into a sb message?

jaiveerk: Oh, looking at the v1 code I think I understand - if I understand correctly, we should have a SenderOptions field that contains a marshaller? And then construct the sender with that

v2/sender/sender.go
jaiveerk: renamed the field to be sbSender as well

.github/workflows/v2-build-and-ut.yaml
```diff
@@ -18,6 +18,11 @@ jobs:
         with:
           go-version: 1.18
 
+      - name: golangci-lint
```
serbrech, line 21: 👍 , but split this out in a separate PR please

v2/sender.go
```diff
@@ -0,0 +1,106 @@
...
+
+var _ CustomSender = &Sender{}
+
+// NewSender takes in a Sender and a Marshaller to create a new object that can send messages to the ServiceBus queue
+func NewSender(sender SBSender, options SenderOptions) *Sender {
+	return &Sender{sbSender: sender, options: options}
+}
+
+// SendMessage marshals the input struct, runs middleware to modify the returned ServiceBus message, and uses the Sender to send the message to the ServiceBus queue
+func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options SendOption) error {
```
serbrech, line 45: `options ...SendOptions`

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func (s SenderOptions) Marshaller() Marshaller {
+	return s.marshaller
+}
+
+// SetTypeHandler sets the ServiceBus message's type to the original struct's type
+func SetTypeHandler(mb MessageBody) func(ctx context.Context, msg *azservicebus.Message) error {
```
serbrech, line 69: SetMessageType

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+
+// SetMessageIdHandler sets the ServiceBus message's ID to a user-specified value
+func SetMessageIdHandler(messageId *string) func(ctx context.Context, msg *azservicebus.Message) error {
+	return func(ctx context.Context, msg *azservicebus.Message) error {
+		msg.MessageID = messageId
+		return nil
+	}
+}
+
+func GetMessageType(mb MessageBody) string {
```
serbrech, line 86: private

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s SenderOptions) Marshaller() Marshaller {
+	return s.marshaller
+}
+
+// SetTypeHandler sets the ServiceBus message's type to the original struct's type
+func SetTypeHandler(mb MessageBody) func(ctx context.Context, msg *azservicebus.Message) error {
+	return func(ctx context.Context, msg *azservicebus.Message) error {
+		msgType := GetMessageType(mb)
+		msg.ContentType = &msgType
```
serbrech, line 72: the message type should be set on a custom property, not on the ContentType

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
+)
+
+// MessageBody is a type to represent that an input message body can be of any type
+type MessageBody any
+
+// CustomSender is a wrapper around the ServiceBus sender that allows for users to introduce middleware to modify the ServiceBus message before it's sent
+type CustomSender interface {
+	SendMessage(ctx context.Context, mb MessageBody, options ...func(ctx context.Context, msg *azservicebus.Message) error) error
```
serbrech, line 16: remove context from the option func

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// SBSender is satisfied by *azservicebus.Sender
+type SBSender interface {
```
serbrech, line 20: MessageSender

serbrech: or `AzServiceBusSender`

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type SenderOptions struct {
+	marshaller Marshaller
+}
+
+var _ CustomSender = &Sender{}
+
+// NewSender takes in a Sender and a Marshaller to create a new object that can send messages to the ServiceBus queue
+func NewSender(sender SBSender, options SenderOptions) *Sender {
```
serbrech, line 37: pointer

serbrech: if options nil, provide default value

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// run user-provided middleware
+	for _, option := range options {
+		err := option(ctx, msg)
+		if err != nil {
+			return fmt.Errorf("failed to run middleware: %s", err)
+		}
+	}
+
+	err = d.sbSender.SendMessage(ctx, msg, &azservicebus.SendMessageOptions{}) // sendMessageOptions currently does nothing
```
serbrech, line 57: single line err handling

v2/sender.go
```diff
@@ -0,0 +1,96 @@
...
+	// uses a marshaller to marshal the message into a service bus message
+	msg, err := d.options.Marshaller().Marshal(mb)
+	if err != nil {
+		return fmt.Errorf("failed to marshal original struct into ServiceBus message: %s", err)
+	}
+
+	// run user-provided middleware
+	for _, option := range options {
+		err := option(ctx, msg)
+		if err != nil {
```
serbrech, line 52: single line err handling

serbrech: fmt.Errorf : use %w for error wrapping, not %s

v2/sender.go
jaiveerk: moving this to be inside the Send function like we discussed, removing the option func entirely

v2/sender.go
```diff
@@ -0,0 +1,122 @@
...
+	msgTypeField = "type"
+)
+
+// MessageBody is a type to represent that an input message body can be of any type
+type MessageBody any
+
+// CustomSender is a wrapper around the ServiceBus sender that allows for users to introduce middleware to modify the ServiceBus message before it's sent
+type CustomSender interface {
+	SendMessage(ctx context.Context, mb MessageBody, options ...func(msg *azservicebus.Message) error) error
+}
```
serbrech, line 22: drop this interface, this can be defined user side if they want to.
we shouldn't provide interfaces

v2/sender.go
```diff
@@ -0,0 +1,122 @@
...
+func NewSender(sender AzServiceBusSender, options *SenderOptions) *Sender {
+	if options == nil {
+		options = &SenderOptions{marshaller: &DefaultJSONMarshaller{}}
+	}
+	return &Sender{sbSender: sender, options: options}
+}
+
+// SendMessage marshals the input struct, runs middleware to modify the returned ServiceBus message, and uses the Sender to send the message to the ServiceBus queue
+func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options ...func(msg *azservicebus.Message) error) error {
+	// uses a marshaller to marshal the message into a service bus message
```
serbrech, line 51: unecessary comment, repeats the code. remove it

v2/sender.go
```diff
@@ -0,0 +1,122 @@
...
+	msg, err := d.options.Marshaller().Marshal(mb)
+	if err != nil {
+		return fmt.Errorf("failed to marshal original struct into ServiceBus message: %w", err)
+	}
+
+	// set type in application properties
```
serbrech, line 57: same. remove comment

v2/sender.go
```diff
@@ -0,0 +1,122 @@
...
+	msgType := getMessageType(mb)
+	msg.ApplicationProperties = map[string]interface{}{msgTypeField: msgType}
+
+	// run user-provided msg options
```
serbrech, line 61: and same here :)

serbrech (approved):
LGTM

PR #92, opened 2022-12-20
Added linter CI

jaiveerk:


.github/workflows/v2-build-and-ut.yaml
```diff
@@ -27,3 +27,9 @@ jobs:
         run: |
           cd v2
           go test -v ./...
+
+      - name: golangci-lint
```
serbrech, line 31: Should it be run in a different job to run in parallel with build/test?

serbrech (approved):

PR #98, opened 2023-01-10
add OnFailure and OnComplete hook

serbrech:
Adding OnFailure and OnSuccess hook function to ManagedSettling handler.
This allows for example to report QoS on message handling via special log sync.

added unit tests for the handler

fixed 2 bugs
- return after deadletterring was missing
- standartd MaxAttemptRetryDecision had the comparison reversed

### DIFF
--- a/v2/managedsettling.go
+++ b/v2/managedsettling.go
@@ -27,9 +27,11 @@ func (m *ManagedSettler) Handle(ctx context.Context, settler MessageSettler, mes
 	}
 	if err := settler.CompleteMessage(ctx, message, nil); err != nil {
 		log(ctx, err)
+		m.options.OnFailure(ctx, message, err)
 		// if we fail to complete the message, we log the error and let the message lock expire.
 		// we cannot do more at this point.
 	}
+	m.options.OnComplete(ctx, message)
 }
 
 // RetryDecision allows to provide custom retry decision.
@@ -47,7 +49,7 @@ type MaxAttemptsRetryDecision struct {
 }
 
 func (d *MaxAttemptsRetryDecision) CanRetry(_ error, message *azservicebus.ReceivedMessage) bool {
-	return message.DeliveryCount > d.MaxAttempts
+	return message.DeliveryCount < d.MaxAttempts
 }
 
 // RetryDelayStrategy can be implemented to provide custom delay retry strategies.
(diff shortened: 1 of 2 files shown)


v2/managedsettling.go
```diff
@@ -27,9 +27,11 @@ func (m *ManagedSettler) Handle(ctx context.Context, settler MessageSettler, mes
 	}
 	if err := settler.CompleteMessage(ctx, message, nil); err != nil {
 		log(ctx, err)
+		m.options.OnFailure(ctx, message, err)
```
jaiveerk, line 30: should we have the same OnFailure func to run after deadlettering AND after completing the message? I feel like a failure to call settler.CompleteMessage is a different kind of error/a sign of an internal issue as opposed to whatever may arise from failing to handle the message within the user-provided `ManagedSettlingFunc`

v2/managedsettling.go
```diff
@@ -113,11 +129,15 @@ func (m *ManagedSettler) handleError(ctx context.Context, settler MessageSettler
 			Reason:             to.Ptr(handleErr.Error()),
 			PropertiesToModify: nil,
 		})
+		// this could be a special hook to have more control on deadlettering, but keeping it simple for now
```
jaiveerk, line 132: Do you think it's a good idea to deadletter any time we hit failure and have run out of retries? What if when we see that the message has reached its retry limit, we use the Abandon settlement?

Also, correct me if I'm wrong, but we're eventually going to provide a default set of settlers to interface with the queue directly, right? So that users don't have to write their own settlements? I forgot how exactly that plays into the current structure we have now

v2/managedsettling.go
```diff
@@ -27,9 +27,11 @@ func (m *ManagedSettler) Handle(ctx context.Context, settler MessageSettler, mes
 	}
 	if err := settler.CompleteMessage(ctx, message, nil); err != nil {
 		log(ctx, err)
+		m.options.OnFailure(ctx, message, err)
 		// if we fail to complete the message, we log the error and let the message lock expire.
 		// we cannot do more at this point.
 	}
+	m.options.OnComplete(ctx, message)
```
jaiveerk, line 34: do we want to have onComplete just call the complete settler by default?

jaiveerk (changes requested):

v2/managedsettling.go
serbrech: I think that's enough at the moment yes. the user can still infer the deadletter because he controls the canretry func, and can inspect the message. in our usecase in fleet I don't feel like we need more at this point

v2/managedsettling.go
serbrech: no, because that's a hook for the user to do stuff like logging. not to modify the behavior of the settling entirely. 
for that you can implement your own handler.

v2/managedsettling.go
serbrech: abandon requeues the message.. so that would basically retry the message when we run out of retry.
the user can set a MaxDelivery on the queue or subscription lower than the MaxRetry option and get the behavior you describe.
this is opioninated, moving to the deadletter from the application so that we can log it/track it, as opposed to letting SB moving it to the DLQ, which is invisible from the application.

jaiveerk: that makes sense, that we should let the user track it and handle it at the application level

jaiveerk (approved):

jaiveerk (approved):

jaiveerk (approved):

jaiveerk (approved):

PR #100, opened 2023-01-19
make marshaller field settable

pdaru:
bug-fix: making Marshaller field settable so user can provide their own marshallers.


serbrech:
lgtm

serbrech (approved):

PR #111, opened 2023-03-08
Enable OTel tracing propagation in go-shuttle/v2

wenxuan0923:
Add OTel tracing propagation to go-shuttle v2. This is because we recently changed AKS-rp frontend and async to use Opentelemetry for tracing instead of Opencensus. See linked PRs for details:

https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/7705767
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/7720051


v2/tracing/tracing.go
```diff
@@ -5,57 +5,80 @@ import (
...
+
+const (
+	traceCarrierField = "traceCarrier"
+	serviceTracerName = "go-shuttle"
 )
 
+var tracer trace.Tracer
+
+func init() {
+	tp := tracesdk.NewTracerProvider(tracesdk.WithSampler(tracesdk.AlwaysSample()))
```
serbrech, line 24: we are in a library here. shouldn't this come from the current settings of the service that uses this library?

v2/tracing/tracing.go
```diff
@@ -5,57 +5,80 @@ import (
...
+	return ctx, span
 }
 
-// GetKeyValues implements tab.Carrier interface
-func (mw *messageWrapper) GetKeyValues() map[string]interface{} {
-	if mw.message != nil {
-		return mw.message.ApplicationProperties
+// This method will extract the remote trace context from the traceCarrier if exists,
+// and inject it into the passed in context. If the traceCarrier doesn't contain a
+// valid trace context, the passed in context will be returned as is.
```
serbrech, line 76: are you sure this is the right behavior?
I was expecting that a new traceContext would be added to the span? but that may be wrong.

v2/sender.go
```diff
@@ -103,6 +104,19 @@ func SetMessageTTL(ttl time.Duration) func(msg *azservicebus.Message) error {
 	}
 }
 
+// SetTraceCarrier set the trace carrier in the message's ApplicationProperties
+// it is for tracing propagation through the ServiceBus queue messages
+func SetTraceCarrier(traceCarrier map[string]string) func(msg *azservicebus.Message) error {
```
serbrech, line 109: I'm not sure how you expect the sender to use this.

When you send a message, the trace data is in the context right?
let's split this PR in 2.
the handler looks better. I think the sender needs to be transparent. this requires too much from the user.

v2/tracing/tracing.go
serbrech: are there any examples on how libraries allow to integrate with services that uses opentelemetry for tracing?

v2/tracing/tracing_test.go
```diff
@@ -0,0 +1,74 @@
...
+		message         *azservicebus.ReceivedMessage
+		carryRemoteSpan bool
+	}{
+		{
+			description:     "nil message, should start a new span without parent",
+			message:         nil,
+			carryRemoteSpan: false,
+		},
+		{
+			description: "should set span attributes from message",
```
serbrech, line 36: could you actually verify that the span attributes do have the expected values in your test?

v2/sender.go
```diff
@@ -103,6 +104,15 @@ func SetMessageTTL(ttl time.Duration) func(msg *azservicebus.Message) error {
 	}
 }
 
+// SetMessageTrace inject the spanContext inside the input context into a service bus message
+func SetMessageTrace(ctx context.Context) func(msg *azservicebus.Message) error {
```
serbrech, line 108: could you move this func to the tracing package?

so to explicitely propagate the trace on message send, it will look like

`sender.Send(ctx, msg, tracing.PropagateFrom(ctx))`

v2/tracing/tracing.go
serbrech: From the otel implementation, you are right:

// The returned Context will be a copy of ctx and contain the extracted
// tracecontext as the remote SpanContext. If the extracted tracecontext is
// invalid, the passed ctx will be returned directly instead.

resolving

serbrech (approved):

PR #135, opened 2023-07-17
[fix] Stop lock-renewer when we know we can

serbrech:
The current lock-renewer is taking a relaxed approach to deciding when to stop attempting to renew. 
It exits the renewal loop on context cancellation only. all other errors are logged, but do not cause the renewer to stop.
This relies on the fact that the processor creates a separate child context per message, so when the message processing is done, the context will eventually be disposed of, causing the matching lock renewer to stop as well.

downside:
- it leaves a window where the handler has sent a message disposition (abandoned, completed, etc..) and has **not** exited yet. In these cases, the renewer can call Renew and get back an error because the lock is lost.
- if a serverside event causes unrecoverable errors, the lock renewer would also rely on context cancelation.

This PR handles this better by explicitely stopping the renewer goroutine when:

- a message has completed its handling
- The error returned is un-recoverable for the target message (LockLost/Auth)

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -1,4 +1,4 @@
-go 1.18
+go 1.19
 
 module github.com/Azure/go-shuttle/v2
 
--- a/v2/processor_test.go
+++ b/v2/processor_test.go
@@ -85,14 +85,11 @@ func TestProcessorStart_ContextCanceledAfterStart(t *testing.T) {
 			ReceiveInterval: to.Ptr(1 * time.Second),
 		})
 	ctx, cancel := context.WithCancel(context.Background())
-	var err error
-	go func() { err = processor.Start(ctx) }()
+	errCh := make(chan error)
+	go func() { errCh <- processor.Start(ctx) }()
 	cancel()
 	g := NewWithT(t)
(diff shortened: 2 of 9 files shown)


v2/go.mod
```diff
@@ -1,4 +1,4 @@
-go 1.18
+go 1.19
```
serbrech, line 1: use of atomic.Bool

v2/processor_test.go
```diff
@@ -85,14 +85,11 @@ func TestProcessorStart_ContextCanceledAfterStart(t *testing.T) {
...
-	go func() { err = processor.Start(ctx) }()
+	errCh := make(chan error)
+	go func() { errCh <- processor.Start(ctx) }()
 	cancel()
 	g := NewWithT(t)
-	g.Eventually(func(g Gomega) {
-		g.Expect(err).ToNot(BeNil())
-	}).Should(Succeed())
-	g.Expect(err).To(Equal(context.Canceled))
+	g.Eventually(errCh).Should(Receive(Equal(context.Canceled)))
```
serbrech, line 92: fix a race detected by go test -race

karenychen (approved):

PR #153, opened 2023-08-17
fix: respecting context cancelation since sdk does not

imiller31:
Adding explicit context handling since azure sdk has shown to not respect timeouts

### DIFF
--- a/v2/sender.go
+++ b/v2/sender.go
@@ -65,10 +65,23 @@ func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options ...fun
 		ctx, cancel = context.WithTimeout(ctx, d.options.SendTimeout)
 		defer cancel()
 	}
-	if err := d.sbSender.SendMessage(ctx, msg, nil); err != nil { // sendMessageOptions currently does nothing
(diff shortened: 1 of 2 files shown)


v2/sender.go
```diff
@@ -65,10 +65,23 @@ func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options ...fun
 		ctx, cancel = context.WithTimeout(ctx, d.options.SendTimeout)
 		defer cancel()
 	}
-	if err := d.sbSender.SendMessage(ctx, msg, nil); err != nil { // sendMessageOptions currently does nothing
```
karenychen, line 68: maybe add back this comment since we are passing in nil?

v2/sender.go
```diff
@@ -120,7 +133,22 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 		return fmt.Errorf("failed to send message batch: %w", err)
```
karenychen, line 133: this should be deleted?

karenychen:
Adding @serbrech, @imiller31 and @erinborders for viz

serbrech (approved):

PR #154, opened 2023-08-18
Add publisher prometheus metrics

karenychen:


v2/sender.go
```diff
@@ -43,14 +45,14 @@ type SenderOptions struct {
 }
 
 // NewSender takes in a Sender and a Marshaller to create a new object that can send messages to the ServiceBus queue
-func NewSender(sender AzServiceBusSender, options *SenderOptions) *Sender {
+func NewSender(sender AzServiceBusSender, entityName string, options *SenderOptions) *Sender {
```
serbrech, line 48: I think the name should either be dropped, or moved into SenderOptions and named appropriately to indicate it is used for metrics.
then there is no breaking change and it's optional.

if there is no option, you just set it to something like "default"

serbrech: just drop it. as it's simpler IMO. we can add it later if needed.

serbrech (approved):

PR #178, opened 2023-11-07
Allow custom extract function in NewTracingHandler

minhng22:
Allow customer to pass in custom extract function when creating new tracing handler with NewTracingHandler(..)

This is necessary since we want more manipulation on attributes of the span that is created in the Extract function.


v2/tracing.go
```diff
@@ -5,13 +5,19 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
 	"github.com/Azure/go-shuttle/v2/otel"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // NewTracingHandler is a shuttle middleware that extracts the context from the message Application property if available,
 // or from the existing context if not, and starts a span.
-func NewTracingHandler(next Handler) HandlerFunc {
+func NewTracingHandler(next Handler,
+	extractFn func(ctx context.Context, message *azservicebus.ReceivedMessage) (context.Context, trace.Span),
```
serbrech, line 14: this makes it a breaking change in the interface.

let's use a variadic option parameter instead, so that it does not break current consumers.

serbrech: that will require to create a TracingHandler type (and not just a func) so that the option can be applied onto something.

```go
type TracingHandler struct {
    EnrichSpan func(context.Context, trace.Span)
}
```

```go
func NewTracingHandler(next Handler, options ...func(t *TracingHandler)) HandlerFunc {
   t := &TracingHandler{}
   for _, opt in range := options {
       opt(t)
   }
   return t.Handle
}
```

then you can provide this helper:

```go
func WithSpanEnrichement(enrich func(context.Context, trace.Span) func(t *TracingHandler)) {
    return func(t *TracingHandler) {
       t.EnrichSpan = enrich
    }
}
```

minhng22: updated

v2/tracing.go
```diff
@@ -5,13 +5,18 @@ import (
...
 // or from the existing context if not, and starts a span.
-func NewTracingHandler(next Handler) HandlerFunc {
+func NewTracingHandler(next Handler,
+	extractFuncs ...func(ctx context.Context, message *azservicebus.ReceivedMessage) (context.Context, trace.Span)) HandlerFunc {
+	if len(extractFuncs) == 0 {
+		extractFuncs = append(extractFuncs, otel.Extract)
+	}
 	return func(ctx context.Context, settler MessageSettler, message *azservicebus.ReceivedMessage) {
-		ctx, span := otel.Extract(ctx, message)
+		ctx, span := extractFuncs[len(extractFuncs)-1](ctx, message)
```
minhng22, line 19: Why getting the last item in `extractFuncs[]`: From the current usage, I am not sure if creating a new span for each `extractFunc` in `extractFuncs` is a good idea, plus using `defer` in for loop is discouraged, so the overall logic will be changed more than I'd like. There is also no use case yet to pass in multiple `extractFunc`, so I decided to just stick to current logic and just get the last `extractFunc`.

serbrech: >I am not sure if creating a new span for each extractFunc in extractFuncs is a good idea

definitely not wanted.
My suggestion was to allow the user to enrich the span that is already extracted. We do not want to let the consumer control _how_ the span is _extracted_ from the message, because that would require them to also know how the span is _propagated_ (where it is stored on the message).

exposing an enrich func allows the framework to control the span extraction, but lets the user pull additional information to enrich the span, which will then cascade down in the trace.

v2/tracing_test.go
```diff
@@ -68,6 +70,15 @@ func Test_NewTracingHandler(t *testing.T) {
 			hasTraceContextOnContext: true,
 			hasTraceContextOnMessage: true,
 		},
+		{
+			name:                     "has traceCtx on context - custom extractFn",
+			hasTraceContextOnContext: true,
+			hasTraceContextOnMessage: false,
+			extractFn: func(ctx context.Context, message *azservicebus.ReceivedMessage) (context.Context, trace.Span) {
+				ctx, span := gootel.Tracer("test-tracer").Start(ctx, "test-span")
```
serbrech, line 78: your example is taking the trace from the receiver context.
but not extracting the context fron the message anymore, so you lose the trace propagation from sender to receiver.

v2/tracing.go
serbrech: notice how the signature of the `enrich` func argument takes in a trace.Span?
this is because we let goshuttle control how the span is extracted, and only allow the user to enrich it with whatever attributes they need.

minhng22: updated in the suggested direction.

v2/tracing_test.go
```diff
@@ -68,6 +69,14 @@ func Test_NewTracingHandler(t *testing.T) {
...
 			hasTraceContextOnMessage: true,
 		},
+		{
+			name:                     "has traceCtx on message - extract from message- custom span name",
+			hasTraceContextOnContext: true,
+			hasTraceContextOnMessage: true,
+			spanEnrichFunc: func(c context.Context, t trace.Span) {
+				t.SetName("custom-span-name")
+			},
+		},
```
minhng22, line 79: Not sure how to test that span name has been set, or how we can test our enrichment in general. Maybe we should use a `ReadWriteSpan` instead...

minhng22: cmt outdated

minhng22:
Addendum to description of change:
* Refactor tracing handler logic
* Simplify otel.Extract(..)

serbrech (approved):

PR #192, opened 2024-01-03
Split logs into different levels

karenychen:
Adapt slog and split logs into different levels

### DIFF
--- a/v2/logging.go
+++ b/v2/logging.go
@@ -12,9 +12,9 @@ func getLogger() *slog.Logger {
 	return log
 }
 
-// EnableHandler allows to set a custom slog.Handler to be used by the go-shuttle logger.
+// SetLogHandler allows to set a custom slog.Handler to be used by the go-shuttle logger.
 // If handler is nil, the default slog logger will be used.
-func EnableHandler(handler slog.Handler) {
+func SetLogHandler(handler slog.Handler) {
 	if handler == nil {
 		log = slog.Default()
 	} else {
--- a/v2/logging.go
+++ b/v2/logging.go
@@ -12,8 +12,12 @@ func getLogger() *slog.Logger {
 	return log
 }
 
-func SetSlogHandler(handler slog.Handler) {
-	if handler != nil {
+// EnableHandler allows to set a custom slog.Handler to be used by the go-shuttle logger.
+// If handler is nil, the default slog logger will be used.
+func EnableHandler(handler slog.Handler) {
+	if handler == nil {
+		log = slog.Default()
+	} else {
 		log = slog.New(handler)
 	}
 }
--- a/v2/logging.go
+++ b/v2/logging.go
@@ -1,23 +1,50 @@
 package shuttle
 
 import (
-	"io"
+	"context"
 	"log/slog"
 )
 
-// logging is disabled by default
-var log = slog.New(slog.NewTextHandler(io.Discard, nil))
+type Logger interface {
+	Info(s string)
+	Warn(s string)
+	Error(s string)
+}
 
-func getLogger() *slog.Logger {
-	return log
+var getLogger = func(ctx context.Context) Logger {
+	return &contextLogger{ctx: ctx, logger: slog.Default()}
+}
+
+// Deprecated: Use SetLogHandler instead to adapt slog.
+// SetLoggerFunc sets the function to be used to acquire a logger when go-shuttle logs.
+func SetLoggerFunc(fn func(ctx context.Context) Logger) {
+	getLogger = fn
 }
 
 // SetLogHandler allows to set a custom slog.Handler to be used by the go-shuttle logger.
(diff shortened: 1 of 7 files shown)


v2/logging.go
```diff
@@ -2,44 +2,30 @@ package shuttle
 
 import (
 	"context"
-	"fmt"
-	"os"
-	"time"
+
+	"github.com/sirupsen/logrus"
```
serbrech, line 6: I did not take a dependency on a logger on purpose.
what benefit does this provide?
the exported Logger interface already supported levels.
none of the additional levels are used from what I can see?

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
-func log(ctx context.Context, a ...any) {
-	if os.Getenv("GOSHUTTLE_LOG") == "ALL" {
-		l := getLogger(ctx)
-		if l == nil {
-			return
-		}
-		l.Info(fmt.Sprint(a...))
+// EnableHandler allows to set a custom slog.Handler to be used by the go-shuttle logger.
+// If handler is nil, the default slog logger will be used.
+func EnableHandler(handler slog.Handler) {
```
serbrech, line 17: This func name isn't working from a package perspective

`SetLogHandler` ?

karenychen: hm might have been a bad commit... updated the name to `SetLogHandler`

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
 import (
-	"context"
-	"fmt"
-	"os"
-	"time"
+	"io"
+	"log/slog"
 )
 
-type Logger interface {
```
serbrech, line 10: since the interface that we expose and use for logging has Info, Warn and Error, can we keep it for now, and hook the slog specific func behind it as default?

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
-	Info(s string)
-	Warn(s string)
-	Error(s string)
-}
-
-// SetLoggerFunc sets the function to be used to acquire a logger when go-shuttle logs.
-func SetLoggerFunc(fn func(ctx context.Context) Logger) {
```
serbrech, line 17: We can deprecate this and keep it around.
maybe even **move** the func into a "deprecated.go" file

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
-
-func (l *printLogger) Warn(s string) {
-	fmt.Println(append(append([]any{}, "WARN - ", time.Now().UTC(), " - "), s)...)
-}
+// logging is disabled by default
+var log = slog.New(slog.NewTextHandler(io.Discard, nil))
 
-func (l *printLogger) Error(s string) {
-	fmt.Println(append(append([]any{}, "ERROR - ", time.Now().UTC(), " - "), s)...)
+func getLogger() *slog.Logger {
```
serbrech, line 11: to keep the same interface, this needs to return a small adapter to slog.Logger to implement the shuttle.Logger interface from L10

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
+	return log
 }
 
-func log(ctx context.Context, a ...any) {
-	if os.Getenv("GOSHUTTLE_LOG") == "ALL" {
-		l := getLogger(ctx)
-		if l == nil {
-			return
-		}
-		l.Info(fmt.Sprint(a...))
```
serbrech, line 43: This can be removed, and replaced everywhere by getlogger().Info/Warn/Error

v2/logging.go
```diff
@@ -1,45 +1,23 @@
...
-	"fmt"
-	"os"
-	"time"
+	"io"
+	"log/slog"
 )
 
-type Logger interface {
-	Info(s string)
-	Warn(s string)
```
serbrech, line 12: these funcs don't take context.
2 ways around it: 
- this is used internally, but implemented by consumers. we can create a new ContextLogger interface that take in the ctx, and use the ContextLogger everywhere in goshuttle
- make getLogger take in a ctx, and use that context in the internal calls. that requires a bit of care on our side (to get a new logger when needed, and not reuse one when the context is different) but I think that's quite easy in go-shuttle

v2/logging.go
karenychen: added back Logger interface

v2/logging.go
```diff
@@ -1,23 +1,50 @@
...
-		log = slog.New(handler)
+		handler = slog.Default().Handler()
 	}
+	getLogger = func(ctx context.Context) Logger {
+		return &contextLogger{ctx: ctx, logger: slog.New(handler)}
+	}
+}
+
+type contextLogger struct {
+	ctx    context.Context
```
serbrech, line 36: add a comment to acknowledge that storing the ctx is bad practice, and tha it will go away when we finish deprecating the previous log configuration

serbrech (approved):

PR #197, opened 2024-03-05
support scheduled msgs from azservicebus sdk

pdaru:


serbrech:
Thanks for the PR Pankil.
we already support scheduling message via the SetScheduledAt and Delay sender options here:

https://github.com/Azure/go-shuttle/blob/cd911547445076b15101d23068e2fe039e9e829b/v2/sender.go#L188C1-L203C2

for batch handling with different schedule on each message, the messages can be converted to servbusmessage with options using https://github.com/Azure/go-shuttle/blob/cd911547445076b15101d23068e2fe039e9e829b/v2/sender.go#L101C18-L101C37
the list of message can then be sent as a batch.

if the sequence number was needed, I think I'd look into modifying the options instead of whole new funcs.

I'll leave the PR open in case others need this. we can look into it then.

PR #198, opened 2024-03-05
Bump azservicebus to v1.6.1

karenychen:
bump azservicebus to v1.6.1. this version includes a fix to the slow sending problem: https://github.com/Azure/azure-sdk-for-go/issues/21684


imiller31 (approved):
LGTM, excited to consume this fix

PR #210, opened 2024-04-04
[metrics] Add Consecutive Connection Check Metrics

karenychen:
Add consecutive connection check Metrics for the SB connectivity probe

### DIFF
--- a/v2/metrics/common/utils.go
+++ b/v2/metrics/common/utils.go
@@ -1,0 +1,41 @@
+package common
+
+import (
+	prom "github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+func HasLabel(m *dto.Metric, key string, value string) bool {
+	for _, pair := range m.Label {
+		if pair == nil {
+			continue
+		}
+		if pair.GetName() == key && pair.GetValue() == value {
+			return true
+		}
+	}
+	return false
+}
+
+func HasLabels(m *dto.Metric, labels map[string]string) bool {
+	for key, value := range labels {
+		if !HasLabel(m, key, value) {
+			return false
+		}
+	}
+	return true
+}
+
+// collect calls the function for each metric associated with the Collector
+func Collect(col prom.Collector, do func(*dto.Metric)) {
+	c := make(chan prom.Metric)
+	go func(c chan prom.Metric) {
+		col.Collect(c)
+		close(c)
+	}(c)
+	for x := range c { // eg range across distinct label vector values
+		m := &dto.Metric{}
+		_ = x.Write(m)
+		do(m)
+	}
+}
--- a/v2/metrics/processor/types.go
+++ b/v2/metrics/processor/types.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
+	"github.com/Azure/go-shuttle/v2/metrics/common"
 	prom "github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 )
@@ -14,6 +15,9 @@ const (
 	messageTypeLabel   = "messageType"
 	deliveryCountLabel = "deliveryCount"
 	successLabel       = "success"
+	namespaceLabel     = "namespace"
+	entityLabel        = "entity"
+	subscriptionLabel  = "subscription"
 )
 
 var (
@@ -49,6 +53,11 @@ func newRegistry() *Registry {
 			Help:      "number of messages being handled concurrently",
 			Subsystem: subsystem,
 		}, []string{messageTypeLabel}),
+		ConsecutiveConnectionCount: prom.NewGaugeVec(prom.GaugeOpts{
+			Name:      "receiver_consecutive_connection_count",
+			Help:      "number of consecutive connection successes or failures",
+			Subsystem: subsystem,
+		}, []string{namespaceLabel, entityLabel, subscriptionLabel, successLabel}),
 	}
 }
 
@@ -65,7 +74,8 @@ func (m *Registry) Init(reg prom.Registerer) {
 		m.MessageHandledCount,
 		m.MessageLockRenewedCount,
 		m.MessageDeadlineReachedCount,
-		m.ConcurrentMessageCount)
+		m.ConcurrentMessageCount,
+		m.ConsecutiveConnectionCount)
 }
 
 type Registry struct {
@@ -74,6 +84,7 @@ type Registry struct {
 	MessageLockRenewedCount     *prom.CounterVec
 	MessageDeadlineReachedCount *prom.CounterVec
 	ConcurrentMessageCount      *prom.GaugeVec
+	ConsecutiveConnectionCount  *prom.GaugeVec
 }
 
 // Recorder allows to initialize the metric registry and increase/decrease the registered metrics at runtime.
@@ -86,6 +97,8 @@ type Recorder interface {
 	IncMessageHandled(msg *azservicebus.ReceivedMessage)
 	IncMessageReceived(float64)
 	IncConcurrentMessageCount(msg *azservicebus.ReceivedMessage)
+	IncConsecutiveConnectionSuccessCount(namespace, entity, subscription string)
+	IncConsecutiveConnectionFailureCount(namespace, entity, subscription string)
 }
 
 // IncMessageLockRenewedSuccess increase the message lock renewal success counter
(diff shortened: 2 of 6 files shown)


serbrech:
curious how usasble this will be, considering that you also need to take into account the prometheus scraping interval?
or do you plan to use this via the informer in a liveness probe?

karenychen:
This will tell us the current health status of any namespace. I am planning to use this with ADXMon to alert on namespace unavailable

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+
+// HealthChecker performs periodic health checks on the Service Bus Senders and Receivers.
+// It uses azservicebus.Sender.NewMessageBatch() and azservicebus.Receiver.PeekMessages() to perform the health checks.
+type HealthChecker struct {
+	// clients is a map of namespace name to azservicebus.Client.
+	clients map[string]*azservicebus.Client
+	// entity is the name of the queue or topic.
+	entity string
+	// subscription is the name of the subscription. Leave empty for queues and senders.
+	subscription string
```
serbrech, line 22: if the subscription is optional most of the time (only for receiver, and only for topics), should this be in the option structs too?
How does this work for topics anyways, we can't ensure that all subscribers understand this "protocol" so we cannot safely send a heathcheck message onto the topic. all subscriptions will receive it.

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+		subscription: subscription,
+		interval:     interval,
+		options:      options,
+	}
+}
+
+// StartSenderPeriodicHealthCheck starts a periodic health check for the sender.
+// It uses azservicebus.Sender.NewMessageBatch() to perform the health check.
+// Stops when the context is cancelled.
+func (h *HealthChecker) StartSenderPeriodicHealthCheck(ctx context.Context) {
```
serbrech, line 57: the Processor Start func blocks.

> Start starts processing on all the receivers of the processor and blocks until all processors are stopped

I think we should stay consistent and let the consumer decide to move it to a background goroutine.

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+	} else {
+		r, err = client.NewReceiverForSubscription(h.entity, h.subscription, nil)
+	}
+	if err != nil {
+		processor.Metric.IncHealthCheckFailureCount(namespace, h.entity, h.subscription)
+		return err
+	}
+	sbCtx, cancel := context.WithTimeout(ctx, h.options.HealthCheckTimeout)
+	defer cancel()
+	_, err = r.PeekMessages(sbCtx, 1, nil)
```
serbrech, line 120: does Peek blocks until it gets 1 message?
if there is no message, does it return an error?

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
+	"github.com/Azure/go-shuttle/v2/metrics/processor"
+	"github.com/Azure/go-shuttle/v2/metrics/sender"
+)
+
+type HealthCheckFunc func(ctx context.Context, namespace string, client *azservicebus.Client) error
+
+// HealthChecker performs periodic health checks on the Service Bus Senders and Receivers.
+// It uses azservicebus.Sender.NewMessageBatch() and azservicebus.Receiver.PeekMessages() to perform the health checks.
```
serbrech, line 15: improve the doc comment to specify when each is used.

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+func (h *HealthChecker) senderHealthCheck(ctx context.Context, namespace string, client *azservicebus.Client) error {
+	s, err := client.NewSender(h.entity, nil)
+	if err != nil {
+		sender.Metric.IncHealthCheckFailureCount(namespace, h.entity)
+		return err
+	}
+	sbCtx, cancel := context.WithTimeout(ctx, h.options.HealthCheckTimeout)
+	defer cancel()
+	_, err = s.NewMessageBatch(sbCtx, nil)
+	if err != nil {
```
serbrech, line 98: single line err handling

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+		sender.Metric.IncHealthCheckFailureCount(namespace, h.entity)
```
serbrech, line 99: can we move the metric handling iton the calling func? or is that then making the select too messy?

v2/health.go
```diff
@@ -0,0 +1,127 @@
...
+func (h *HealthChecker) receiverHealthCheck(ctx context.Context, namespace string, client *azservicebus.Client) error {
+	var r *azservicebus.Receiver
+	var err error
+	if h.subscription == "" {
+		r, err = client.NewReceiverForQueue(h.entity, nil)
+	} else {
+		r, err = client.NewReceiverForSubscription(h.entity, h.subscription, nil)
+	}
+	if err != nil {
+		processor.Metric.IncHealthCheckFailureCount(namespace, h.entity, h.subscription)
```
serbrech, line 115: should the metric differentiate sender/receiver health check?

v2/health.go
karenychen: no message available doesn't return an error: https://github.com/Azure/azure-sdk-for-go/blob/d2f6a15a733221e831e16b7223366017e72b2875/sdk/messaging/azservicebus/internal/mgmt.go#L138-L141

v2/health.go
karenychen: The reason why we start each client's `periodicHealthCheck`  in a goroutine is because we might have more than one namespace. If each `periodicHealthCheck` is blocking, we will never be able to move past the first namespace.

v2/health.go
karenychen: This metric is mainly used for alerting/monitoring the SB namespace health. Distinction between a sender and a receiver health alert can help us quickly determine which side is having issues and improve our TTU (time to understand).

v2/health.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+	if _, err = sbSender.NewMessageBatch(ctx, nil); err != nil {
+		return err
+	}
+	return sbSender.Close(ctx)
+}
+
+// IncHealthCheckMetric increments the sender health check metric based on the health check result.
+func (s *SenderHealthChecker) IncHealthCheckMetric(namespaceName string, healthCheckErr error) {
```
serbrech, line 104: I think this func should stay internal. unless you have a use-case for exporting it

v2/health.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	// note: PeekMessages() does not return an error when the entity is empty
+	if _, err = sbReceiver.PeekMessages(ctx, 1, nil); err != nil {
+		return err
+	}
+	return sbReceiver.Close(ctx)
+}
+
+// IncHealthCheckMetric increments the receiver health check metric based on the health check result.
+func (r *ReceiverHealthChecker) IncHealthCheckMetric(namespaceName string, healthCheckErr error) {
```
serbrech, line 133: ditto

v2/health.go
karenychen: My original thought is this allows applications to define their own implementation of `HealthCheckable`. For example, if they want to use a different method to verify the connection, or if they want to increment some of their own metrics.

Also open to keeping it internal.

v2/health.go
```diff
@@ -0,0 +1,146 @@
...
+// ReceiverHealthChecker performs health checks on azservicebus.Receiver.
+type ReceiverHealthChecker struct {
+	EntityName       string
+	SubscriptionName string
+}
+
+// HealthCheck performs a health check on the azservicebus.Receiver by peeking a message.
+func (r *ReceiverHealthChecker) HealthCheck(ctx context.Context, client *azservicebus.Client, cancelFunc context.CancelFunc) error {
+	defer cancelFunc()
+	sbReceiver, err := r.createReceiver(client)
```
serbrech, line 121: should we do 

```golang
defer func () { 
  if err := sbReceiver.Close(ctx); err!=nil {
    getLogger(ctx).Errorf(err)
  }
}()
```
just after the create?

wondering about the consequences of the Peek failing, and leaving a receiver not closed.
from the outside, if you get an error, you can't close the receiver.

serbrech: so you need to define a var err outside the defer, and you an still modify its value inside the defer, and have it returned.
see https://stackoverflow.com/a/19934989/156415 for example

so rather than my simplified example:
```golang
var err error
var sbReceiver ...
sbReceiver, err = r.createReceiver(client)
defer func () { 
  err = sbReceiver.Close(ctx); 
}()
return err
```

serbrech: same for the sender case

v2/health.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// SenderHealthChecker performs health checks on azservicebus.Sender.
+type SenderHealthChecker struct {
+	EntityName string
+}
+
+// HealthCheck performs a health check on azservicebus.Sender by creating a new message batch.
+func (s *SenderHealthChecker) HealthCheck(ctx context.Context, client *azservicebus.Client, cancelFunc context.CancelFunc) error {
+	defer cancelFunc()
```
serbrech, line 92: It is unusual to pass down a context and a cancelfunc. I'd consider that a smell.
I think in this case, you can call defer cancel from the caller and it's going to be same.

serbrech (approved):

karenychen:
note: confirmed the metrics to be working 
image

PR #215, opened 2024-04-15
Enable Configuration of Processor.Start() Retries

karenychen:
This PR enables retries on `Processor.Start()`. Users can configure this with `StartMaxAttempt` and `StartRetryDelayStrategy`. `Start()` returns the last error encountered while starting the processor

### DIFF
--- a/v2/processor.go
+++ b/v2/processor.go
@@ -84,7 +84,7 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
 		if options.ReceiveInterval != nil {
 			opts.ReceiveInterval = options.ReceiveInterval
 		}
-		if options.MaxConcurrency > 0 {
+		if options.MaxConcurrency >= 0 {
 			opts.MaxConcurrency = options.MaxConcurrency
 		}
 		if options.StartMaxAttempt > 0 {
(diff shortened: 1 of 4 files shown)


v2/processor.go
```diff
@@ -74,7 +103,31 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
 }
 
 // Start starts the processor and blocks until an error occurs or the context is canceled.
+// It will retry starting the processor based on the StartMaxAttempt and StartRetryDelayStrategy.
+// Returns the last error encountered while starting the processor.
 func (p *Processor) Start(ctx context.Context) error {
+	var savedError error
+	for attempt := 1; attempt <= p.options.StartMaxAttempt; attempt++ {
```
imiller31, line 110: Nit: Is there a reason we are starting at 1 instead of 0?

karenychen: initially put down 1 for better readability during the early exit block. changed back to start the loop from 0

v2/processor.go
```diff
@@ -74,7 +103,31 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
 }
 
 // Start starts the processor and blocks until an error occurs or the context is canceled.
+// It will retry starting the processor based on the StartMaxAttempt and StartRetryDelayStrategy.
+// Returns the last error encountered while starting the processor.
 func (p *Processor) Start(ctx context.Context) error {
+	var savedError error
+	for attempt := 0; attempt < p.options.StartMaxAttempt; attempt++ {
+		if err := p.start(ctx); err != nil {
+			savedError = err
```
serbrech, line 112: use errors.Join and return them all.

serbrech (changes requested):
small changes requested

serbrech (approved):
LGTM!

PR #217, opened 2024-04-16
Enable MultiProcessor in processor.go

karenychen:
User can add more than one receiver to listen to with go-shuttle's Processor.

Use `NewMultiProcessor()` to initialize a list of receivers in a Processor

Note: max concurrency count is shared across all receivers.

MultiProcessor is not yet compatible with LockRenewer.


v2/processor.go
```diff
@@ -115,16 +148,17 @@ func (p *Processor) Start(ctx context.Context) error {
...
+	receiver := p.receivers[index]
+	log(ctx, fmt.Sprintf("starting processor %d", index))
+	messages, err := receiver.ReceiveMessages(ctx, p.options.MaxConcurrency, nil)
 	if err != nil {
-		return err
+		return fmt.Errorf("processor %d failed to receive messages: %w", index, err)
 	}
-	log(ctx, fmt.Sprintf("received %d messages - initial", len(messages)))
+	log(ctx, fmt.Sprintf("processor %d received %d messages - initial", index, len(messages)))
 	processor.Metric.IncMessageReceived(float64(len(messages)))
```
imiller31, line 159: wonder if we need to have different metrics for the different receivers

karenychen: the main challenge here is go-shuttle doesn't record namespace or entity name. It will be difficult to label the receivers.

v2/processor.go
```diff
@@ -82,23 +83,55 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
...
-		receiver:          receiver,
+		receivers:         []Receiver{receiver},
 		handle:            handler,
 		options:           opts,
 		concurrencyTokens: make(chan struct{}, opts.MaxConcurrency),
 	}
 }
 
-// Start starts the processor and blocks until an error occurs or the context is canceled.
+func (p *Processor) AddReceiver(receivers ...Receiver) {
```
serbrech, line 93: problem with this API is that you don't control the order in which it's called.
it it allowed to add a receiver after calling Start?

I would prefer making this part of the initialization of the Processor

v2/processor.go
serbrech: you have now "labeled" them with an index.
if you want to identify them more specifically, we'd need to have a way to know the namespace or entity that the receiver is configured with.
or allow the user to give each receiver a name.

that could get rid of that index. you can then store the receivers in a map with the name as a key.
not certain how much better it is :)

v2/processor.go
```diff
@@ -82,23 +83,55 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
...
+	}
+	return errors.Join(allErrs...)
+}
+
+// StartAtIndex starts the processor with receiver index and blocks until an error occurs or the context is canceled.
 // It will retry starting the processor based on the StartMaxAttempt and StartRetryDelayStrategy.
 // Returns a combined list of errors during the start attempts or ctx.Err() if the context
 // is cancelled during the retries.
-func (p *Processor) Start(ctx context.Context) error {
+func (p *Processor) StartAtIndex(ctx context.Context, index int) error {
```
serbrech, line 126: why do we need this api? not fond of exposing the index.
how does the user know which receiver is which index? especially with the AddReceiver api?

v2/processor.go
```diff
@@ -82,23 +83,55 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
...
+// It will retry starting the processor based on the StartMaxAttempt and StartRetryDelayStrategy.
+// Returns a combined list of errors encountered during each processor start.
+func (p *Processor) Start(ctx context.Context) error {
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, len(p.receivers))
+	for i := range p.receivers {
+		wg.Add(1)
+		go func(index int) {
+			defer wg.Done()
+			err := p.StartAtIndex(ctx, index)
```
serbrech, line 107: I see, StartAtIndex is only used here.
Let's keep it internal. that lets us change how we organize them

v2/processor.go
karenychen: note: synced offlined and agreed on allowing user to give receiver a name.

v2/processor.go
karenychen: added `NewMultiProcessor()`

v2/processor.go
```diff
@@ -115,16 +177,17 @@ func (p *Processor) Start(ctx context.Context) error {
...
-		return err
+		return fmt.Errorf("processor %s failed to receive messages: %w", receiverName, err)
 	}
-	log(ctx, fmt.Sprintf("received %d messages - initial", len(messages)))
-	processor.Metric.IncMessageReceived(float64(len(messages)))
+	log(ctx, fmt.Sprintf("processor %s received %d messages - initial", receiverName, len(messages)))
+	processor.Metric.IncMessageReceived(receiverName, float64(len(messages)))
 	for _, msg := range messages {
-		p.process(ctx, msg)
+		p.process(ctx, receiverName, msg)
```
serbrech, line 190: I think you should pass down the receiver here

v2/processor.go
```diff
@@ -133,37 +196,38 @@ func (p *Processor) start(ctx context.Context) error {
...
 		msgContext, cancel := context.WithCancel(ctx)
 		// cancel messageContext when we get out of this goroutine
 		defer cancel()
 		defer func() {
 			<-p.concurrencyTokens
-			processor.Metric.IncMessageHandled(message)
+			processor.Metric.IncMessageHandled(receiverName, message)
 			processor.Metric.DecConcurrentMessageCount(message)
 		}()
 		processor.Metric.IncConcurrentMessageCount(message)
```
serbrech, line 229: nit: you could also add the receiver name to the concurrent message count metric.
that allows you to count all, or per receiver, and see the metric change when we failover for example.

v2/processor.go
```diff
@@ -25,6 +26,18 @@ type MessageSettler interface {
 	RenewMessageLock(ctx context.Context, message *azservicebus.ReceivedMessage, options *azservicebus.RenewMessageLockOptions) error
 }
 
+type ReceiverImpl struct { // shuttle.Receiver is already an exported interface
```
serbrech, line 29: Keep this struct private, an dname it just `receiver`.
add a constructor: 

shuttle.NamedReceiver(receiver, name) Receiver {
}

add Name() string to the Receiver interface we have, and always wrap the incoming receiver from the single receiver processor, with a default name.

Then you only depend on the interface.

WDYT?

v2/processor.go
```diff
@@ -25,6 +26,18 @@ type MessageSettler interface {
...
+	name       string
+	sbReceiver Receiver
+}
+
+func NewReceiverImpl(name string, sbReceiver Receiver) *ReceiverImpl {
```
serbrech, line 34: make this return `Receiver` interface

v2/processor.go
serbrech: If you prefer to keep the struct public, which is ok, I have an issue with the `Impl` suffix.

NamedReceiver is better, but I'm worried that it's too precise (what happens when we want to add more properties?)

ReceiverEx (Extended) is a little better than Impl IMO, because it tells it's extended, not that it's an implementation of an interface...

open to your suggestion/preference.

v2/processor.go
```diff
@@ -115,16 +177,17 @@ func (p *Processor) Start(ctx context.Context) error {
 }
 
 // start starts the processor and blocks until an error occurs or the context is canceled.
-func (p *Processor) start(ctx context.Context) error {
-	log(ctx, "starting processor")
-	messages, err := p.receiver.ReceiveMessages(ctx, p.options.MaxConcurrency, nil)
+func (p *Processor) start(ctx context.Context, receiverName string) error {
```
serbrech, line 180: nit: why not taking a ReceiverEx here?

v2/processor.go
```diff
@@ -133,37 +196,38 @@ func (p *Processor) start(ctx context.Context) error {
...
 		}
 	}
-	log(ctx, "exiting processor")
-	return ctx.Err()
+	log(ctx, fmt.Sprintf("exiting processor %s", receiverName))
+	return fmt.Errorf("processor %s stopped: %w", receiverName, ctx.Err())
 }
 
-func (p *Processor) process(ctx context.Context, message *azservicebus.ReceivedMessage) {
+func (p *Processor) process(ctx context.Context, receiverName string, message *azservicebus.ReceivedMessage) {
```
serbrech: take in a ReceiverEx?

v2/processor.go
```diff
@@ -81,24 +94,73 @@ func NewProcessor(receiver Receiver, handler HandlerFunc, options *ProcessorOpti
...
 
-// Start starts the processor and blocks until an error occurs or the context is canceled.
+// Start starts processing on all the receivers of the processor and blocks until all processors are stopped or the context is canceled.
+// It will retry starting the processor based on the StartMaxAttempt and StartRetryDelayStrategy.
+// Returns a combined list of errors encountered during each processor start.
+func (p *Processor) Start(ctx context.Context) error {
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, len(p.receivers))
+	for name := range p.receivers {
+		wg.Add(1)
```
serbrech, line 132: Here, if startOne panics, we're not entering wg.Done() because we don't recover from it.
so I think that the wg.Wait() will keep us stuck forever.

could you try to add a test for that by generating a panic somehow inside startOne ?

serbrech: in terms of fixing it, I've been itching to use conc library in go-shuttle, but unsure whether that's wise as a library.
we should handle the panic I think...

karenychen: added panic handling + UTs

serbrech (approved):
Ok let's get this in. we might want to adjust some things once we write integration code around it.

PR #223, opened 2024-04-24
Enable Sender Failover

karenychen:
Enable `shuttle.Sender` to failover to a new `AzServiceBusSender`. Caller can initiate the failover by calling `sender.FailOver()`

### DIFF
--- a/v2/sender.go
+++ b/v2/sender.go
@@ -82,9 +82,7 @@ func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options ...fun
 
 	errChan := make(chan error)
 	go func() {
-		d.mu.RLock()
-		err := d.sbSender.SendMessage(ctx, msg, nil) // sendMessageOptions currently does nothing
-		d.mu.RUnlock()
+		err := d.sendMessage(ctx, msg, nil) // sendMessageOptions currently does nothing
 		if err != nil {
 			errChan <- fmt.Errorf("failed to send message: %w", err)
 		} else {
@@ -144,9 +142,7 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 		return fmt.Errorf("failed to send message: %w", ctx.Err())
 	}
 
-	d.mu.RLock()
-	batch, err := d.sbSender.NewMessageBatch(ctx, &azservicebus.MessageBatchOptions{})
-	d.mu.RUnlock()
+	batch, err := d.newMessageBatch(ctx, &azservicebus.MessageBatchOptions{})
 	if err != nil {
 		return err
 	}
@@ -164,10 +160,7 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 	errChan := make(chan error)
 
 	go func() {
-		d.mu.RLock()
-		err := d.sbSender.SendMessageBatch(ctx, batch, nil)
-		d.mu.RUnlock()
-		if err != nil {
+		if err := d.sendMessageBatch(ctx, batch, nil); err != nil {
 			errChan <- fmt.Errorf("failed to send message batch: %w", err)
 		} else {
 			errChan <- nil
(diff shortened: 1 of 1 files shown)


v2/sender.go
```diff
@@ -135,7 +139,9 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 	if ctx.Err() != nil {
 		return fmt.Errorf("failed to send message: %w", ctx.Err())
 	}
-	
+
+	d.mu.RLock()
```
serbrech, line 143: Just to be clear, this means that calling FailOver has to wait for messages to be sent before failing over.
This make it quite critical to have a timeout on the Send calls, or you can be deadlocked. (you cannot send, and you cannot failover).
Do I understand the code correctly?

v2/sender.go
```diff
@@ -178,9 +184,17 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 
 // AzSender returns the underlying azservicebus.Sender instance.
 func (d *Sender) AzSender() AzServiceBusSender {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
 	return d.sbSender
 }
 
+func (d *Sender) FailOver(sender AzServiceBusSender) {
+	d.mu.Lock()
```
serbrech, line 193: conversely, this will block all Send calls on the RLock call, which is done before accessing teh sbSender variable, so that's fair.

v2/sender.go
```diff
@@ -74,6 +76,8 @@ func (d *Sender) SendMessage(ctx context.Context, mb MessageBody, options ...fun
 		defer cancel()
 	}
 
+	d.mu.RLock()
+	defer d.mu.RUnlock()
```
serbrech, line 80: Any reason to unlock only after the select?
can't we move this lock/unlock sequence inside the go func() below, so that the lock is taken just in time and released ASAP?

v2/sender.go
```diff
@@ -178,9 +184,17 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 
 // AzSender returns the underlying azservicebus.Sender instance.
 func (d *Sender) AzSender() AzServiceBusSender {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
 	return d.sbSender
 }
 
+func (d *Sender) FailOver(sender AzServiceBusSender) {
```
serbrech, line 192: Add documentation on the Failover, and the impact on the consumers of the Sender object.

v2/sender.go
```diff
@@ -29,6 +30,7 @@ type AzServiceBusSender interface {
 // Sender contains an SBSender used to send the message to the ServiceBus queue and a Marshaller used to marshal any struct into a ServiceBus message
 type Sender struct {
 	sbSender AzServiceBusSender
+	mu       sync.RWMutex
```
serbrech, line 33: document what the mutex is for, when to take RLock and when to take Lock.
I think, in essence, all usages of sbSender should use RLock and ensure to always release the lock.
the only place to use Lock is in the Failover func.

v2/sender.go
```diff
@@ -29,6 +30,7 @@ type AzServiceBusSender interface {
 // Sender contains an SBSender used to send the message to the ServiceBus queue and a Marshaller used to marshal any struct into a ServiceBus message
 type Sender struct {
 	sbSender AzServiceBusSender
```
serbrech, line 32: document that any usage of sbSender must consider the mutex and take the appropriate lock.

serbrech:
Here is an interesting alternative implementation that has pros an cons.
you can reduce the scope of the lock, the number of places that have to care about a lock, at the cost of introducing a channel.
that allows you to control when you effectively attempts a send, and interleave the swap.

I think that the mutex is a good choice in this case though. but I'm curious about how the 2 implementation would differ :)

v2/sender_test.go
serbrech, line 1: Please make sure to run tests that use -race flag, and ensure that the tests we have would uncover a race.
we might need a specific test that does sends in multiple goroutines while failing over.

v2/sender.go
karenychen: Yes. Even without the failover, we should always have a timeout to the Sends because they could hang indefinitely.

v2/sender_test.go
karenychen: Added UT for sending in multiple goroutines while failing over. The sender passes `go test -race`, but the fake lockrenewer has a race error. I will raise that as an issue so we can fix it in a later PR.

v2/sender.go
karenychen: updated to release the lock right after Send()

v2/sender.go
```diff
@@ -154,7 +164,10 @@ func (d *Sender) SendMessageBatch(ctx context.Context, messages []*azservicebus.
 	errChan := make(chan error)
 
 	go func() {
-		if err := d.sbSender.SendMessageBatch(ctx, batch, nil); err != nil {
+		d.mu.RLock()
+		err := d.sbSender.SendMessageBatch(ctx, batch, nil)
+		d.mu.RUnlock()
```
serbrech, line 169: always use defer
if SendMessageBatch panics, you won't be unlocking the mutex, and we're in trouble.
Moving the `lock/defer unlock` inside the `go func()` already scopes it better than before.
You could wrap even further in a func to avoid including the error handling in the critical section.

serbrech (approved):
🚀 !

Issue #241, opened 2024-07-30
SDK Receiver.RenewMessageLock() should pass in a context with timeout


karenychen:
https://github.com/Azure/go-shuttle/blob/f3bc6b375d9bff3a7bf05e54eb4926a7e3d2a147/v2/lockrenewer.go#L130

Receiver.RenewMessageLock() should pass in a context with timeout to prevent hanging

PR #242, opened 2024-07-30
Add option to timeout during LockRenewal calls

karenychen:
Fix for https://github.com/Azure/go-shuttle/issues/241

### DIFF
--- a/v2/lockrenewer.go
+++ b/v2/lockrenewer.go
@@ -182,7 +182,7 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
 
 func (plr *peekLockRenewer) renewMessageLock(ctx context.Context, message *azservicebus.ReceivedMessage, options *azservicebus.RenewMessageLockOptions) error {
 	lockLostErr := &azservicebus.Error{Code: azservicebus.CodeLockLost}
-	renewalTimeout := message.LockedUntil.Sub(time.Now())
+	renewalTimeout := time.Until(*message.LockedUntil)
 	if renewalTimeout < 0 {
 		// if the lock is already expired, we should not attempt to renew it.
 		return lockLostErr
@@ -191,9 +191,9 @@ func (plr *peekLockRenewer) renewMessageLock(ctx context.Context, message *azser
 		renewalTimeout = *plr.renewalTimeout
 	}
 	// we should keep retrying until lock expiry or until message context is done
-	for message.LockedUntil.Sub(time.Now()) > 0 && ctx.Err() == nil {
+	for time.Until(*message.LockedUntil) > 0 && ctx.Err() == nil {
 		err := plr.renewMessageLockWithTimeout(ctx, renewalTimeout, message, options)
-		// exit the renewal if we get any error other than context canceled.
+		// exit the renewal if we get any error other than context canceled or if the error is nil.
 		if !errors.Is(err, context.Canceled) {
 			return err
 		}
@@ -209,5 +209,5 @@ func (plr *peekLockRenewer) renewMessageLockWithTimeout(ctx context.Context, tim
 	getLogger(ctx).Info(fmt.Sprintf("renewing lock with timeout: %s", timeout))
 	renewalCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
-	return plr.renewMessageLock(renewalCtx, message, options)
+	return plr.lockRenewer.RenewMessageLock(renewalCtx, message, options)
 }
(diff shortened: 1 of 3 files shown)


v2/lockrenewer.go
```diff
@@ -170,3 +182,13 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
 		}
 	}
 }
+
+func (plr *peekLockRenewer) renewMessageLock(ctx context.Context, message *azservicebus.ReceivedMessage, options *azservicebus.RenewMessageLockOptions) error {
+	if *plr.renewalTimeout > 0 {
+		getLogger(ctx).Info(fmt.Sprintf("renewing lock with timeout: %s", *plr.renewalTimeout))
+		var cancel func()
+		ctx, cancel = context.WithTimeout(ctx, *plr.renewalTimeout)
```
imiller31, line 190: Might be faster to understand here if you use `:=` with a new ctx variable name like `renewalCtx`

karenychen: using the same variable since we are only setting the timeout if the renewTimeout > 0. If it is set to a negative value, it is disabled and we don't want to mess with that?

imiller31 (approved):

v2/lockrenewer.go
```diff
@@ -34,11 +39,16 @@ type LockRenewalOptions struct {
 // NewRenewLockHandler returns a middleware handler that will renew the lock on the message at the specified interval.
 func NewRenewLockHandler(options *LockRenewalOptions, handler Handler) HandlerFunc {
 	interval := 10 * time.Second
+	lockRenewalTimeout := interval
 	cancelMessageContextOnStop := true
 	metricRecorder := processor.Metric
 	if options != nil {
 		if options.Interval != nil {
 			interval = *options.Interval
+			lockRenewalTimeout = interval
```
serbrech, line 48: I don't mind this but I don't think making the timeout relative to the interval makes much sense.
to reduce network bandwidth and increase throughput, you increase that interval to be less chatty.
so, say 1min lock 20s interval.
I would still want just a few seconds of timeout. not 20s.

karenychen: I think this is more of a default if application code doesn't set anything in its options? Since we know the interval has to be less than the lock duration, this default will allow the lock renewer exit before the lock expires.

karenychen: edit: changed to default to 5s if not set or 0

v2/lockrenewer.go
```diff
@@ -170,3 +179,35 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
 		}
 	}
 }
+
+func (plr *peekLockRenewer) renewMessageLock(ctx context.Context, message *azservicebus.ReceivedMessage, options *azservicebus.RenewMessageLockOptions) error {
+	lockLostErr := &azservicebus.Error{Code: azservicebus.CodeLockLost}
+	renewalTimeout := time.Until(*message.LockedUntil) // assumes messages.LockedUntil is set
```
serbrech, line 185: I think that's fine, but if the middleware is used with Peek, that might crash.
it wouldn't make sense anyways to have this renewer for a message that isn't locked. so I'm fine with crashing here

v2/lockrenewer.go
```diff
@@ -170,3 +179,35 @@ func (plr *peekLockRenewer) startPeriodicRenewal(ctx context.Context, message *a
...
+		return lockLostErr
+	}
+	if *plr.renewalTimeout > 0 {
+		renewalTimeout = *plr.renewalTimeout
+	}
+	// we should keep retrying until lock expiry or until message context is done
+	for time.Until(*message.LockedUntil) > 0 && ctx.Err() == nil {
+		err := plr.renewMessageLockWithTimeout(ctx, renewalTimeout, message, options)
+		// exit the renewal if we get any error other than context canceled or if the error is nil.
+		if !errors.Is(err, context.Canceled) {
```
serbrech, line 197: I think this is too narrow now.
if the error is lockLost, there is no point retrying for example, and we weren't retrying before, but now you will retry N times and get N lockLost error, instead of exiting promptly and canceling the message context.

I would have just explicitely retried on the deadline exceeded (which is our timeout case).

v2/lockrenewer.go
karenychen: added a check here to return lock lost if there is no LockUntil

v2/lockrenewer.go
karenychen: oops, meant to say `errors.Is(err, context.DeadlineExceeded)`.

I think that is what I have? We are only retrying for the timeout case. If the renew fails for any reason other than timeout, we are not retrying

karenychen: added a UT here to show: 
image

serbrech (approved):

PR #243, opened 2024-08-05
[lockrenewer] add logs for each renewal attempt

karenychen:


v2/lockrenewer.go
```diff
@@ -199,10 +198,15 @@ func (plr *peekLockRenewer) renewMessageLock(ctx context.Context, message *azser
...
-		// exit the renewal if we get any error other than context deadline exceeded or if the error is nil.
+		if renewErr != nil {
+			getLogger(ctx).Error(fmt.Sprintf("failed to renew lock: %s", renewErr))
+			span.AddEvent("failed to renew message lock", trace.WithAttributes(attribute.String("errorDetails", renewErr.Error())))
+		}
+		// exit the renewal if no error or if we get any error other than context deadline exceeded
 		if !errors.Is(renewErr, context.DeadlineExceeded) {
 			return renewErr
 		}
+		getLogger(ctx).Error(fmt.Sprintf("renewal error is %s, retrying the renewal fast", renewErr))
```
serbrech, line 209: err is not nil (L201)
err is not DeadlineExceeded (206)
but, it will be nil on L209, and you would get a nil ref exception because you call the Stringer on it?

serbrech: you already always log it on L202, so this is duplicate anyways

serbrech (approved):

PR #276, opened 2025-05-01
[sender] Support sender SetSubject() and SetTo() options

karenychen:
Support sender SetSubject() and SetTo() options

### DIFF
--- a/v2/sender.go
+++ b/v2/sender.go
@@ -266,6 +266,22 @@ func SetMessageTTL(ttl time.Duration) func(msg *azservicebus.Message) error {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


v2/sender.go
```diff
@@ -266,6 +266,22 @@ func SetMessageTTL(ttl time.Duration) func(msg *azservicebus.Message) error {
 	}
 }
 
+// SetSubject sets the ServiceBus message's Subject property to a user-specified value
+func SetSubject(subject string) func(msg *azservicebus.Message) error {
```
serbrech, line 270: are there limits to the size of these properties on the SB/AMQP protocol?

PR #277, opened 2025-05-02
[processor] Set ReceiveInterval to be 0 when we receive less than requested

karenychen:
Set ReceiveInterval to be 0 when we receive less than requested to improve Processor efficiency


serbrech:
I think this is the wrong way around?

it's if we receive a number of message **equal** to what we requested that we can assume there is more on the queue waiting, so we should receive again right away.

if we receive less than the request, then it's likely that there is nothing left on the queue, so we can keep the interval until next receive.

karenychen:
closing this as this is something the client can configure manually

PR #280, opened 2026-06-05
Remove MultiProcessor API

karenychen:
## Summary
- remove the unused NewMultiProcessor API and its ReceiverEx wrapper
- restore Processor to a single receiver while preserving retry, MaxReceiveCount, metrics, and panic recovery behavior
- remove MultiProcessor docs/examples/tests and keep single-processor panic recovery coverage

## Verification
- go test ./... (from v2)
- go vet ./... (from v2)
- git diff --check
- e2e module compiled against local v2 via temporary modfile/replace

PR #281, opened 2026-06-05
Deprecate MultiProcessor API

karenychen:
## Summary
- mark NewMultiProcessor as deprecated ahead of future removal
- mark the companion ReceiverEx and NewReceiverEx APIs as deprecated
- update the README MultiProcessor section to direct users to NewProcessor instead

## Verification
- go test ./... (from v2)
- go vet ./... (from v2)
- git diff --check

Note: PR #280 remains draft for the later removal after a deprecation release is tagged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,8 @@ type ProcessorOptions struct {
 
 ### MultiProcessor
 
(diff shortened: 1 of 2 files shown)


serbrech (approved):
LGTM

PR #288, opened 2026-06-29
Add processor graceful shutdown

karenychen:
## Summary
- Add `Processor.Close(ctx)` to stop receiving and cancel in-flight handlers during shutdown.
- Track in-flight messages so close can abandon the messages currently held by the processor.
- Abandon tracked messages concurrently as best effort while respecting caller context cancellation.
- Keep failed abandon attempts tracked so a later close can retry them.

## Testing
- `go test -run TestInFlightMessages .`
- `go test -race -run 'TestInFlightMessages|TestProcessor' .`\n- `go test ./...`\n- `git diff --check`


karenychen:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance.

